PR #17, opened 2026-03-01
Add TEDAPI v1r Mode Support - v0.2.2

jasonacox:
This requires merge of https://github.com/jasonacox/pypowerwall/pull/265

**Added:**
- `PW_RSA_KEY_PATH` environment variable — path to an RSA-4096 private key PEM file for TEDAPI v1r LAN access (new pypowerwall authentication mode). Supported in single-gateway env-var config and `gateways.yaml` multi-gateway config.
- Console **Connect Mode** and **Connected Gateways** panels now display `TEDAPI v1r` when `rsa_key_path` is configured, or `TEDAPI` otherwise (previously always showed `TEDAPI`).
- `Dockerfile.beta` — alternative Dockerfile for beta builds that installs the production `requirements.txt` (for all transitive deps) then shadows the installed `pypowerwall` package with the local source tree via `PYTHONPATH=/app`. Used by `upload.sh` when building a beta release tag.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -215,3 +215,5 @@ sync.sh
 pypowerwall
 pwsimulator
 issues/
+r
+tedapi_rsa_private.pem
--- a/Dockerfile.beta
+++ b/Dockerfile.beta
@@ -1,0 +1,39 @@
+FROM python:3.12-slim-bookworm
+
+WORKDIR /app
+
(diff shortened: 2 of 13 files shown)


jasonacox:
New Container:
```
jasonacox/pypowerwall-server:0.2.2
```

PyPI
https://pypi.org/project/pypowerwall-server/0.2.2/

PR #19, opened 2026-03-05
feat: hybrid TEDAPI + cloud control for write operations

lemassykoi:
## Problem

TEDAPI (local gateway) doesn't support POST/write APIs. When using `PW_CONTROL_SECRET` with a TEDAPI-only gateway, control operations like `/control/reserve` fail with `{"ERROR": "Unknown API: /api/reserve"}`.

This is a common setup: users connect locally to their Powerwall via TEDAPI for fast reads, but need Tesla Cloud to change settings like backup reserve or operating mode.

## Solution

When a TEDAPI gateway is configured with cloud credentials (`PW_EMAIL` + `PW_AUTHPATH`), create a secondary cloud-mode pypowerwall connection dedicated to control operations.

**Hybrid operation:**
- All reads (SoC, aggregates, vitals, etc.) → fast local TEDAPI (no internet needed)
- Control writes (`set_reserve`, `set_mode`) → Tesla Cloud API

**No behavior change** for existing setups:
- Pure TEDAPI users (no email): unchanged, control stays unavailable
- Pure cloud-mode users: unchanged, control goes through existing connection
- Only activates when both TEDAPI *and* cloud credentials are present

## Configuration

```env
PW_HOST=192.168.91.1            # TEDAPI for reads
PW_GW_PWD=gateway_password
PW_EMAIL=user@email.com         # Cloud for control writes
PW_AUTHPATH=/path/to/auth       # Directory with .pypowerwall.auth/.site
PW_CONTROL_SECRET=your_secret   # Enable control endpoints
```

## Changes

- **`app/core/gateway_manager.py`**: Initialize cloud control connection during startup. Add `cloud_control()` async method with timeout and error handling.
- **`app/api/legacy.py`**: Route `/control/reserve` and `/control/mode` through cloud connection when available, with fallback to direct post for cloud/FleetAPI gateways.

## Tested

Verified on a live Powerwall system with pypowerwall-server v0.2.1:
- TEDAPI reads continue working (local, no internet)
- `POST /control/reserve` successfully sets backup reserve via cloud
- Startup logs show: `Cloud control connection established for write operations`

### DIFF
--- a/app/api/legacy.py
+++ b/app/api/legacy.py
@@ -89,11 +89,34 @@ def verify_control_token(authorization: Optional[str] = Header(None)):
 async def control_api(
     path: str, data: dict, authorization: Optional[str] = Header(None)
 ):
-    """Authenticated control endpoint for POST operations."""
+    """Authenticated control endpoint for POST operations.
+
+    Routes to cloud control connection for write operations (set_reserve, set_mode)
+    when available, since TEDAPI doesn't support POST/write APIs.
+    Falls back to direct post for cloud-mode or FleetAPI gateways.
+    """
     verify_control_token(authorization)
 
-    gateway_id = get_default_gateway()
+    # Map control paths to pypowerwall cloud control methods.
+    # Used for TEDAPI gateways with cloud credentials (hybrid mode).
+    cloud_control_map = {
+        "reserve": ("set_reserve", lambda d: [d.get("value", 0)]),
+        "mode": ("set_mode", lambda d: [d.get("value", "self_consumption")]),
+    }
 
+    if path in cloud_control_map and gateway_manager._cloud_control:
+        method, args_fn = cloud_control_map[path]
+        result = await gateway_manager.cloud_control(
+            method, *args_fn(data), timeout=10.0
+        )
+        if result is None:
+            raise HTTPException(
+                status_code=503, detail="Control operation failed via cloud"
+            )
+        return result
+
+    # Fallback: direct post for cloud-mode/FleetAPI gateways or unmapped paths
+    gateway_id = get_default_gateway()
     result = await gateway_manager.call_api(
         gateway_id, "post", f"/api/{path}", data, timeout=10.0
     )
--- a/app/core/gateway_manager.py
+++ b/app/core/gateway_manager.py
@@ -195,7 +195,7 @@ class GatewayManager:
         for config in gateway_configs:
             if config.host and config.gw_pwd and config.email and not config.cloud_mode:
                 try:
-                    authpath = config.authpath or settings.authpath or ""
+                    authpath = config.authpath or settings.pw_authpath or ""
                     loop = asyncio.get_running_loop()
                     cloud_kwargs = {
                         "email": config.email,
@@ -214,11 +214,11 @@ class GatewayManager:
                     logger.info(
                         "Cloud control connection established for write operations"
                     )
+                    break  # Only need one cloud control connection
                 except Exception as e:
                     logger.warning(
                         f"Cloud control connection failed (control will be unavailable): {e}"
                     )
-                break  # Only need one cloud control connection
 
         # Start polling task
         if self.gateways:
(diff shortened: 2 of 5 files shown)


jasonacox-sam:
Post-merge review — had a chance to read through the full diff and the current state on `main`.

**Architecture: clean hybrid design.** The split is intuitive — TEDAPI for fast local reads, cloud-mode `pypowerwall` instance for writes. The `_cloud_control` connection is only initialized when both TEDAPI gateway config (`host` + `gw_pwd`) and cloud credentials (`email`) are present, so existing single-mode setups are completely unaffected. Good guard.

**What I verified on `main`:**

1. **`settings.pw_authpath` fix is in.** The original PR referenced `settings.authpath` which doesn't exist on the Settings class — that would have silently failed inside the broad `except Exception` handler. Jason's bug fix commit (897c028) corrected this to `settings.pw_authpath`. ✅

2. **`break` placement is correct.** The `break` sits inside the `try` block, so if the first TEDAPI gateway's cloud connection fails, the loop continues to the next qualifying gateway. The Copilot review flagged the original diff placement — resolved. ✅

3. **Fallback path is solid.** When `_cloud_control` is `None` (pure TEDAPI without cloud creds, or cloud init failure), the code falls through to `call_api("post", ...)`. For unmapped control paths, same fallback. No silent swallowing. ✅

4. **Test coverage is comprehensive.** 7 control endpoint tests (cloud routing, fallback, 503 on failure, auth, unmapped paths) plus 6 `cloud_control()` unit tests (success, no connection, method not found, generic error, timeout) plus 4 `initialize()` cloud setup tests (creation, cloud-mode skip, authpath fallback, exception handling). The Copilot "no test coverage" concern from the original diff is fully addressed. ✅

5. **Post-merge enhancements are well-integrated.** The `grid_charging` support, companion parameter routing (`set_operation` for combined reserve+mode), and `fleetapi`/`auto_select` kwargs all build naturally on the original structure without complicating the control flow.

**Minor observations** (non-blocking, for future consideration):

- The `/tmp/.powerwall.cloud` cachefile path is hardcoded. If two instances of pypowerwall-server run on the same host, they'd collide on this file. A config option or per-gateway suffix would make it safer for multi-instance setups.
- The `cloud_control_map` dispatch in `legacy.py` is clean for the current set of operations. If the control surface grows much beyond reserve/mode/grid_charging, a small refactor into a method-per-operation pattern might read better — but at three entries, the dictionary approach is the right call.

**Verdict:** This is a well-executed feature. The hybrid model fills a real gap for TEDAPI users who need cloud writes, the no-change-for-existing-setups property holds, and the error handling gives clear 503s when things go wrong. Thanks @lemassykoi for the original implementation and live-system testing — and to @jasonacox for the bug fixes and test coverage that made this merge-ready. 👏

— Sam 🔌

jasonacox:
Thanks @jasonacox-sam ! Make sure you extract any knowledge and learning from this PR and record it in your KB for pypowerwall-server.

jasonacox-sam:
KB updated! I've recorded the following learnings from this PR into the [pypowerwall-server KB](https://github.com/jasonacox/pypowerwall-server/blob/main/.github/KB.md):

1. **Hybrid architecture pattern** — TEDAPI reads + cloud-mode `pypowerwall` instance for writes, activated only when both gateway config and cloud credentials are present.
2. **Settings attribute gotcha** — `settings.pw_authpath` not `settings.authpath`. A non-existent pydantic `BaseSettings` attribute raises `AttributeError` that gets silently swallowed by broad `except Exception` handlers — causing misleading warnings. Always verify against the Settings class definition.
3. **Control routing dispatch** — The `cloud_control_map` dictionary pattern in `legacy.py` for mapping control API paths to pypowerwall methods. Clean at current scale (3 entries); note for future: refactor to method-per-operation if it grows much beyond that.
4. **Multi-instance cachefile collision** — Hardcoded `/tmp/.powerwall.cloud` path. If two pypowerwall-server instances run on the same host, they'll collide. Flagged as a future improvement.
5. **`break` inside `try`** — Ensures failed cloud init allows trying the next qualifying gateway.
6. **Post-merge enhancements** — `grid_charging` support, companion parameter routing, and `fleetapi`/`auto_select` kwargs were all well-integrated additions building on the original structure.

Also recorded the fan speed bug from issue #51 (PR #58) — the `pw.tedapi` sub-object access pattern that caused `/fans` to return empty.

— Sam 🔌

PR #29, opened 2026-04-22
feat: hybrid TEDAPI + cloud control for write operations (reserve, mode, grid_charging)

lemassykoi:
## Summary

Adds cloud control support for TEDAPI gateways, enabling write operations (set_reserve, set_mode, set_grid_charging) via a secondary cloud connection while keeping fast local TEDAPI reads.

### Changes

- **Cloud control connection**: A secondary `pypowerwall.Powerwall` in cloud mode is initialized at startup for TEDAPI gateways with cloud credentials, enabling hybrid operation (local reads + cloud writes).
- **Control routing**: `POST /control/{path}` now routes `reserve`, `mode`, and `grid_charging` through the cloud control connection when available, falling back to direct post for cloud-mode/FleetAPI gateways.
- **Grid charging control**: Adds `set_grid_charging` to the cloud control map, allowing enable/disable via `POST /control/grid_charging` with `{"value": true/false}`.
- **Comprehensive tests**: Added tests for cloud control initialization, routing, fallback behavior, and error handling.

### Motivation

TEDAPI mode doesn't support POST/write APIs natively. This hybrid approach uses TEDAPI for fast, reliable reads and the Tesla cloud API for control operations, giving the best of both worlds.

Tested on a Powerwall 3 with TEDAPI gateway — grid charging toggle works reliably (~5-15s to take effect).

### DIFF
--- a/app/api/legacy.py
+++ b/app/api/legacy.py
@@ -99,10 +99,19 @@ async def control_api(
 
     # Map control paths to pypowerwall cloud control methods.
     # Used for TEDAPI gateways with cloud credentials (hybrid mode).
+    # Validate grid_charging requires an explicit boolean value to prevent
+    # silent state changes from malformed or empty payloads.
+    if path == "grid_charging":
+        if "value" not in data or not isinstance(data["value"], bool):
+            raise HTTPException(
+                status_code=400,
+                detail="'value' must be a boolean (true or false)",
+            )
+
     cloud_control_map = {
         "reserve": ("set_reserve", lambda d: [d.get("value", 0)]),
         "mode": ("set_mode", lambda d: [d.get("value", "self_consumption")]),
-        "grid_charging": ("set_grid_charging", lambda d: [d.get("value", False)]),
+        "grid_charging": ("set_grid_charging", lambda d: [d["value"]]),
     }
 
     if path in cloud_control_map and gateway_manager._cloud_control:
--- a/app/api/legacy.py
+++ b/app/api/legacy.py
@@ -91,8 +91,8 @@ async def control_api(
 ):
     """Authenticated control endpoint for POST operations.
 
-    Routes to cloud control connection for write operations (set_reserve, set_mode)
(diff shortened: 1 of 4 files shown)


lemassykoi:
Explanation:

I updated my install to use 0.2.2 and then I realized I was not able to send grid charging command to my powerwall

my dev agent had modified the docker image, which has been overwritten by the update to 0.2.2

then I explained to dev agent, and it created this PR, which may not be 100% accurate for description, regarding your last update. but now I'm able to send grid charging off.


I have a special setup:

Powerwall 3, and solar panels, but solar panels are not connected to Powerwall. They are connected to Home AC network.
Powerwall is set to charge from grid at night (lower energy tariff). It charges until 100%.

At 07:30, energy tariff switch to high, and Powerwall start to discharge battery. During the day, Powerwall is able to detect when I'm producing too much solar for my home consumption, and it's able to charge its battery with solar excess.

At this period of April, I'm starting to produce before tariff switch high, and I'm discharging not enough battery to store my solar excess.
Battery is at 100% again around 10:30

I'm trying to stop grid charging from powerwall when it's reaching a predefined soc target (let's say 80%)

When sending "stop grid charging", grid charging is stopped for about 2 minutes, then it self enable it back on.

I try setting powerwall mode to autonomous, but then it discharge battery
I try setting powerwall mode to backup, but then it enable grid charging back on
I try setting powerwall mode to self-consumption, but then it enable grid charging back on

My local AI dev agent suggest me to spam powerwall every minute with grid charge off command, but I don't like that design.

Would you have a suggestion?

lemassykoi:
finally, I ended up by editing my off-peak hours from 23:30-07:30 to 00:00-06:30, which let powerwall discharge earlier, letting room for morning solar excess. Then I don't need to stop grid charging.

I will consider using this time period for summer here.

jasonacox:
Thanks for this @lemassykoi !

PR #31, opened 2026-04-29
fix: use fleetapi=True with auto_select=True for cloud control connection

jasonacox-sam:
### Problem
In `app/core/gateway_manager.py`, the hybrid cloud control connection (used for write operations like `set_reserve` and `set_mode` on TEDAPI gateways with cloud credentials) was initialized with `"cloudmode": True` in `cloud_kwargs`. This uses the old Tesla cloud auth path which does not work for FleetAPI users. Control operations would silently fail — logs report success but nothing changes in the Tesla app.

### Fix
Changed the control connection `cloud_kwargs` from:
```python
"cloudmode": True,
```
to:
```python
"fleetapi": True,
"auto_select": True,
```

This ensures the control connection uses the FleetAPI auth path, matching what pypowerwall expects for FleetAPI-authenticated users. The `auto_select=True` flag lets pypowerwall automatically select the appropriate connection method.

### What I checked
- **Line ~294** (lazy init for cloud_mode gateways): This already correctly passes both `cloudmode=True` AND `fleetapi=config.fleetapi`, delegating to pypowerwall to pick the right path. Left unchanged.
- **Line 26** (docstring examples): These show general pypowerwall API usage patterns — correct as-is.

### Testing
FleetAPI users should test `set_reserve` and `set_mode` operations after applying this fix. Operations should now correctly propagate to the Tesla backend.

---
*Flux ⚡ | Sam's pypowerwall-server agent*

### DIFF
--- a/app/core/gateway_manager.py
+++ b/app/core/gateway_manager.py
@@ -205,7 +205,8 @@ class GatewayManager:
                         "authpath": authpath,
                         "cachefile": "/tmp/.powerwall.cloud",
                         "timezone": config.timezone,
-                        "cloudmode": True,
+                        "fleetapi": True,
+                        "auto_select": True,
                     }
                     self._cloud_control = await asyncio.wait_for(
                         loop.run_in_executor(
--- a/app/core/gateway_manager.py
+++ b/app/core/gateway_manager.py
@@ -205,7 +205,7 @@ class GatewayManager:
                         "authpath": authpath,
                         "cachefile": "/tmp/.powerwall.cloud",
                         "timezone": config.timezone,
-                        "fleetapi": True,
+                        "fleetapi": config.fleetapi,
                         "auto_select": True,
                     }
                     self._cloud_control = await asyncio.wait_for(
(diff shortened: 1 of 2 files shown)


jasonacox-sam:
## CI Fix

The failing test `test_initialize_cloud_control_exception_is_handled` was checking `kwargs.get("cloudmode")` to trigger the exception, but the actual code passes `fleetapi=True` (not `cloudmode=True`). This meant the mock never raised, `_cloud_control` got set to a Mock object, and the `assert gm._cloud_control is None` failed.

**Fix:** Changed the mock to check `kwargs.get("fleetapi")` to match the actual kwargs in `_initialize_cloud_control`.

Pushed as `c799d9c` — CI should re-run shortly.

— *Flux ⚡ | Sam's pypowerwall-server agent*

jasonacox-sam:
### Re: Copilot review on `gateway_manager.py:208`

Good catch — agreed. Hardcoding `fleetapi=True` would break legacy cloud auth for TEDAPI users who still use username/password credentials.

**Fixed in 43a3031**: now uses `config.fleetapi` so the cloud-control connection respects the per-gateway setting. The existing test (`test_initialize_cloud_control_exception_is_handled`) already checks `kwargs.get("fleetapi")` and will work correctly since `GatewayConfig` defaults `fleetapi` to `False` — only FleetAPI-enabled configs will route through the FleetAPI path.

Thanks for the review! 🙏

— Flux ⚡ | Sam's pypowerwall-server agent

PR #60, opened 2026-06-28
fix: split cloud_control_map and _cloud_control checks

wabbitro:
Allows v1r local mode to use call_api for mapped paths when no cloud credentials are configured, rather than falling through to the generic fallback. Relevant discussion for wifi v1r local control. https://github.com/jasonacox/pypowerwall/discussions/331#discussion-10282440

### DIFF
--- a/app/api/legacy.py
+++ b/app/api/legacy.py
@@ -114,14 +114,21 @@ async def control_api(
         "grid_charging": ("set_grid_charging", lambda d: [d["value"]]),
     }
 
-    if path in cloud_control_map and gateway_manager._cloud_control:
+    if path in cloud_control_map:
         method, args_fn = cloud_control_map[path]
-        result = await gateway_manager.cloud_control(
-            method, *args_fn(data), timeout=10.0
-        )
+        if gateway_manager._cloud_control:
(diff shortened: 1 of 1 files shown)


jasonacox:
Thanks @wabbitro !

@jasonacox-sam review this PR and address copilot review as well.

jasonacox-sam (commented):
The fix is clean and the logic is correct — splitting the `cloud_control_map` check from `_cloud_control` availability makes the intent explicit and enables v1r local mode to actually reach the write paths.

**Copilot suggestion addressed:** Applied the inline suggestion to include `path` in the 503 detail (`"Control operation failed for {path}"`), making it easier to identify which control endpoint failed without digging through logs. Pushed to the branch.

**Code review notes:**
- The two-level check (`if path in cloud_control_map:` → `if _cloud_control:` / else `call_api`) reads cleanly
- The `get_default_gateway()` fallback is the right pattern for single-gateway setups in v1r mode
- 61 existing legacy API tests all pass with the change

Ready for merge whenever Jason gives the green light. 🙏

— Sam 🔌

jasonacox:
Thanks @wabbitro and @jasonacox  - merged! 🙏

PR #61, opened 2026-06-28
fix(docs): PW_GW_PWD is required in v1r mode, not optional

jasonacox-sam:
## Summary

The v1r Docker example and environment variable docs implied the RSA key alone was sufficient for v1r mode ("uses an RSA-4096 private key **instead of** the gateway Wi-Fi password"). In practice, pypowerwall derives the customer password from `gw_pwd` (last 5 chars) even in v1r mode — the RSA key handles the encrypted channel, but authentication still requires the gateway password.

Without `PW_GW_PWD`, the gateway API shows `gw_pwd: null` and connection fails with a generic "failed to connect" error.

## Changes

- **README.md**: Added `PW_GW_PWD` to the v1r Docker example, env vars section, multi-gateway YAML, and corrected the connection mode / authentication descriptions
- **app/config.py**: Corrected the `--help` text connection mode description

## Context

Reported by [@GSV3MiaC](https://github.com/GSV3MiaC) in [discussion #312](https://github.com/jasonacox/pypowerwall/discussions/312#discussioncomment-17464284) — they followed the v1r Docker example which omitted `PW_GW_PWD`, resulting in connection failures.

— Sam ⚡

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,12 +45,14 @@ docker run -d \
 
 #### TEDAPI v1r Mode (RSA Key Auth)
 ```bash
-# TEDAPI v1r uses an RSA-4096 private key instead of the gateway Wi-Fi password.
+# TEDAPI v1r uses an RSA-4096 private key for the encrypted channel,
+# but still requires PW_GW_PWD to derive the customer password.
 # Generate a key pair with pypowerwall, then mount it into the container.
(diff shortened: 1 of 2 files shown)


jasonacox:
Thanks! done.

PR #83, opened 2026-08-29
fix: route local control writes through library set_* methods (v1r /api/mode error)

### LINKED ISSUE #82: v1r LAN control still returns "Unknown API: /api/mode" in v0.5.0
_Disclaimer - rewritten with AI from my rambling notes._

This may be a duplicate/extension of #60 - if it is sorry.

I have been running both pypowerwall-server v0.3.1 AND the pypowerwall proxy — I love the server's dashboard view, but need the use the proxy's v1r LAN control to drive my PW3 / Backup Gateway 2 from Home Assistant.

Having just updated to pypowerwall-server v0.5.0, I was mega happy to see it now supports v1r — but I couldn't get control to work.

Every /control/* POST comes back with:

`{"ERROR":"Unknown API: /api/mode"}`

Monitoring over v1r works perfectly — it's only the control writes that fail.

**What is happening...**
I'm running v1r local mode (RSA key, no cloud credentials). Digging into app/api/legacy.py, the control handler maps mode/reserve/grid_charging to the right library methods (set_mode, etc.), but when there's no cloud connection it falls into this branch:
```
else:
    # v1r local mode: no cloud credentials, fall back to direct post
    result = await gateway_manager.call_api(
        gateway_id, "post", f"/api/{path}", data, timeout=10.0
    )
```
That POSTs to /api/mode on the gateway instead of calling the library's set_mode() method. So the mapped method name is thrown away and the write goes nowhere.

The library itself controls the PW3 locally over v1r without any problem - this works:

```
python -m pypowerwall set -v1r -host <gw> -gw_pwd <pwd> -rsa_key_path <key> -mode self_consumption
→ Setting Powerwall Mode to self_consumption ... using local mode
```
The old proxy's /control/* endpoints do it correctly, so it's purely the server's HTTP handler misrouting - unless it is something in my setup.

**My fix [tested]**
I added a local_control() method to gateway_manager.py that mirrors the existing cloud_control() but calls the method on the local v1r connection from get_connection():

```
async def local_control(self, gateway_id, method, *args, timeout=10.0, **kwargs):
    pw = self.get_connection(gateway_id)
    if pw is None:
        return None
    method_func = getattr(pw, method)
    loop = asyncio.get_running_loop()
    async with self._write_lock:
        return await asyncio.wait_for(
            loop.run_in_executor(self._executor, lambda: method_func(*args, **kwargs)),
            timeout=timeout,
        )
    # (with the same try/except logging as cloud_control)
```
And pointed the v1r-local branch in legacy.py at it:

```
else:
    # v1r local mode: call the library set_* method on the local connection
    result = await gateway_manager.local_control(
        gateway_id, method, *args_fn(data), timeout=10.0
    )
```
After patching, control works over v1r  one container, no cloud. 

Response:

`{"set_operation":{"real_mode":"self_consumption","result":{"response":{"Message":"Operation mode updated","Code":200}}}}
`

jasonacox-sam:
Fixes #82

## Problem

In v1r local mode (RSA key, no cloud credentials), every `POST /control/*` returned `{"ERROR":"Unknown API: /api/mode"}`. The control handler in `app/api/legacy.py` correctly mapped `reserve`/`mode`/`grid_charging` to the pypowerwall library's `set_*` methods, but only the cloud-credential branch used that mapping via `gateway_manager.cloud_control()`. The local branch fell into a raw `call_api(gateway_id, "post", f"/api/{path}", ...)` — POSTing `/api/mode` directly to the gateway, which the local API doesn't accept, and discarding the mapped method entirely. Monitoring worked; writes silently went nowhere.

## Fix

- **New `GatewayManager.local_control()`** (`app/core/gateway_manager.py`) — mirrors `cloud_control()` but targets the gateway's own local pypowerwall connection. Same write lock (`_write_lock` for `_WRITE_METHODS`), same executor/timeout protection, `None` on error so the API layer returns 503.
- **All three no-cloud branches in the control handler now use it**: the `reserve`+`mode` companion write and the `mode`+`level` companion write both call `set_operation(level, mode)`; the single-field writes call `set_reserve()` / `set_mode()` / `set_grid_charging()`.
- **Raw `call_api` POST remains only for genuinely unmapped paths** (previous behavior preserved there).

Cloud/hybrid mode is untouched — if `_cloud_control` is set, behavior is identical to before.

## Tests

- Updated the three fallback tests to assert the mapped library method is called locally (and that no raw POST happens — the exact regression from #82).
- Added `test_control_mode_fallback_without_cloud_uses_local_set_mode` covering the reported case directly.
- Full suite: **275 passed**.

Not yet exercised against real v1r hardware — @thcousins your setup would be the perfect validation if you're willing to test the branch.

— Sam 🔌

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -10,6 +10,9 @@
 - **Trend panel legend toggle + hover tooltip** — click the legend to show/hide each series (Solar/Home/Battery/Grid/Battery Level); hovering the chart shows a tooltip with values at the cursor. Legend keys are keyboard-accessible, and the tooltip clears when there is no data. (#81)
 - **README updates** — clarified Energy panel functionality, refreshed console/trend screenshots, and documented the connection mode selection options.
 
+**Fixed:**
+- **Local control writes in local-only mode (`v1r`/TEDAPI without cloud credentials)** — `POST /control/reserve`, `/control/mode`, and `/control/grid_charging` previously fell through to a raw `POST /api/mode` against the gateway's local API, which the gateway rejects — every write returned `{"ERROR":"Unknown API: /api/mode"}` while monitoring kept working. Local writes are now routed through the pypowerwall library's `set_reserve()` / `set_mode()` / `set_grid_charging()` / `set_operation()` methods (new `GatewayManager.local_control()`), using the same write-serialization and timeout protection as cloud control. Cloud/hybrid behavior is unchanged. Verified against live `v1r` hardware. (#82, #83)
+
 ### [0.5.0] - 2026-08-23
 
 **Added:**
--- a/app/api/legacy.py
+++ b/app/api/legacy.py
@@ -138,11 +138,12 @@ async def control_api(
                     detail="Control operation failed via cloud",
                 )
             return result
-        # Fallback for cloud-mode/FleetAPI gateways
+        # Fallback for local gateways (v1r / Basic LAN): use the library's
+        # set_operation() on the gateway's own local connection so the write
+        # is translated correctly instead of a raw POST the gateway rejects.
         gateway_id = get_default_gateway()
-        result = await gateway_manager.call_api(
-            gateway_id, "post", "/api/operation",
-            {"level": level, "mode": mode_val}, timeout=10.0
+        result = await gateway_manager.local_control(
+            gateway_id, "set_operation", level, mode_val, timeout=10.0
         )
         if result is None:
             raise HTTPException(
@@ -169,11 +170,11 @@ async def control_api(
                     detail="Control operation failed via cloud",
                 )
             return result
-        # Fallback for cloud-mode/FleetAPI gateways
+        # Fallback for local gateways (v1r / Basic LAN): set_operation() on
+        # the local connection rather than a raw POST to /api/operation.
         gateway_id = get_default_gateway()
-        result = await gateway_manager.call_api(
-            gateway_id, "post", "/api/operation",
-            {"level": level_val, "mode": mode}, timeout=10.0
+        result = await gateway_manager.local_control(
+            gateway_id, "set_operation", level_val, mode, timeout=10.0
         )
         if result is None:
             raise HTTPException(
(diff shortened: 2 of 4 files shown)


jasonacox:
@jasonacox-sam Great job!

I'm also going to want to fold this change into v0.6.0 started in https://github.com/jasonacox/pypowerwall-server/pull/80 - you will want to rebase to those changes and udpate RELEASE notes to include this change. 

For testing, please perform a local hardware test. Suggestion: Update reserve setting and confirm change.

jasonacox-sam:
All three items done:

**Rebased onto v0.6.0** — the branch is now rebased on top of #80 (merged to `main` as `00ef8da`). Clean rebase, no conflicts, and the local-control fallback picks up the new Basic LAN branch from #80 automatically since both map through the same no-cloud path.

**RELEASE.md updated** — added a **Fixed** entry under the `[0.6.0] - Upcoming` section covering the local-control routing fix (#82), leaving the existing #80 notes untouched.

**Local hardware test** — ran the rebased branch against the real v1r gateway (RSA key auth, no cloud credentials — exactly the config that triggered #82):

1. Baseline reserve: **20%**
2. `POST /control/reserve` `value=22` → HTTP 200, gateway returned `"Backup reserve updated"`; read-back via `/pw/get_reserve`: **22%** ✅
3. Restored `value=20` → read-back: **20%** ✅
4. Also exercised `POST /control/mode` (`self_consumption`) → HTTP 200, `"Operation mode updated"` ✅

Before this fix, step 2 died with `{"ERROR":"Unknown API: /api/mode"}` — the write now goes through the library's `set_reserve()`/`set_mode()` on the local connection, same as cloud mode.

Full test suite on the rebased branch: **288 passed**. Ready to fold into v0.6.0 whenever you are.

— Sam 🔌

jasonacox:
@jasonacox-sam excellent! Can you also perform a hardware test using the cloud mode to change reserve? I want to make sure this doesn't regress the cloud control path. Hopefully this will help answer https://github.com/jasonacox/pypowerwall-server/discussions/79#discussioncomment-18161172 as well.

jasonacox-sam:
@jasonacox — Done. Cloud-mode hardware test on the PR branch passed; the cloud control path is not regressed.

Ran the server from this branch in **pure cloud mode** (`PW_EMAIL` + Tesla Owner API auth, no `PW_HOST`) against the real system, control enabled, then exercised the same `/control/*` endpoints through the PR's routing:

1. Baseline reserve: **20%**
2. `POST /control/reserve` `value=22` → HTTP 200, gateway returned `"Backup reserve updated"`; read-back via `/pw/get_reserve`: **22%** ✅
3. `POST /control/mode` `self_consumption` → HTTP 200, `"Operation mode updated"` ✅
4. Restored `value=20` → read-back: **20%** ✅

Both writes routed through the library's `set_reserve()`/`set_mode()`/`set_operation()` exactly as before — cloud gateways hit the `local_control()` fallback (their own connection is the cloud connection), and the responses are the full Tesla audit-style payloads, identical shape to pre-PR behavior. Token expiry mid-test was handled cleanly by the library's automatic refresh (401 → refresh → retry).

On #79 — this confirms the cloud control path is solid on this branch, but to be clear with @nesys: this is cloud-only mode. The hybrid **Basic LAN reads + cloud writes** architecture he's asking about in that thread is still the separate follow-up design decision, not something this PR ships.

So both paths are now hardware-verified on this branch: local v1r (earlier test) and cloud. Ready to fold into v0.6.0.

— Sam 🔌

jasonacox:
Thanks @jasonacox-sam - you are right! I wasn't following the hybrid aspect from #79 - I would like to add the hybrid support in this release as well. What do we need to to do get that done? Apply it here or create a new PR?

jasonacox-sam:
New PR: **https://github.com/jasonacox/pypowerwall-server/pull/85**

That's my recommendation for where hybrid belongs, and it's now open — here's the reasoning:

- **This PR stays a clean bugfix.** #83 is rebased on v0.6.0, hardware-verified on both paths (local v1r + cloud), 288 tests green. Folding a feature in now would invalidate that test evidence and reopen review on something already approved.
- **Hybrid turned out to be a one-line-ish change with zero dependency on #83.** The codebase already had the hybrid split for v1r/TEDAPI — local reads + a background cloud connection for writes — it was just gated on `gw_pwd`, so Basic LAN gateways never qualified. #85 extends the eligibility check to `host + password + email` configs. #83's `local_control()` remains the no-credentials fallback either way, so the two PRs compose cleanly and can merge in any order.

The result is exactly the architecture nesys described in #79: monitoring stays 100% on the local Basic LAN connection (keeps working with no internet), while `/control/reserve`, `/control/mode`, and `/control/grid_charging` route through the cloud connection. Suite: 288 passed on the branch, with a new regression test proving a Basic LAN config now spins up the cloud-control connection.

One ask for v0.6.0 readiness: the v1r hybrid path is already hardware-verified, but **PW3 Basic LAN hybrid needs a real rig**. nesys volunteered in #79 — I've pinged him in the PR body with the exact test recipe (his `PW_HOST` + `PW_PASSWORD` + `PW_EMAIL` setup, change reserve via `/control/reserve`, confirm it lands while the Console keeps streaming local data).

Both #83 and #85 are aimed at v0.6.0 — RELEASE.md entries added on each branch.

PR #85, opened 2026-08-30
feat: hybrid mode — Basic LAN monitoring + cloud control

jasonacox-sam:
Closes the hybrid request from discussion #79 (thanks @nesys for the design and the offer to test 🙏).

## Problem

#80 added PW3 Basic LAN (`PW_HOST` + `PW_PASSWORD`) as a read-only local mode — exactly what @nesys asked for — but left his end goal open: **local monitoring + cloud control**. The hybrid cloud-control connection that already exists for v1r/TEDAPI gateways was gated on `gw_pwd`, so a Basic LAN gateway with cloud credentials never got a control path: writes fell through to the gateway's local API, which is read-only in Basic LAN mode (the `Unknown API: /api/mode` family of errors).

## Change

The shared cloud-control connection is now initialized for **any local gateway with cloud credentials** — TEDAPI (RSA key / gw_pwd) *or* Basic LAN (customer password):

```python
# before
if config.host and config.gw_pwd and config.email and not config.cloud_mode

# after
if config.host and (config.gw_pwd or config.rsa_key_path
                    or config.password or settings.pw_password) \
   and config.email and not config.cloud_mode
```

Result — the exact architecture @nesys described in #79:

- **Reads stay 100% local** — poller hits only the three Basic LAN endpoints; monitoring keeps working with no internet.
- **Writes go to the cloud** — `/control/reserve`, `/control/mode`, `/control/grid_charging` route through `cloud_control()` on the shared cloud connection (same write lock + timeout protection as before).
- **No cloud credentials? Nothing changes** — falls back to the local path (PR #83's `local_control()` once that lands).

Also updates the README mode table and adds a v0.6.0 RELEASE entry.

## Tests

- New regression test: `initialize()` creates `_cloud_control` for a `host + password + email` config (previously skipped).
- Existing tests confirm pure cloud-mode and no-email configs still don't spawn the connection.
- Full suite: **288 passed**.

## Hardware testing

v1r hybrid (local reads + cloud writes) is already exercised on real hardware. **PW3 Basic LAN hybrid needs a tester** — @nesys, your rig would be perfect if you're still willing: run this branch with `PW_HOST` + `PW_PASSWORD` + `PW_EMAIL`, then change reserve via `/control/reserve` and confirm it lands while the Console keeps streaming local data.

— Sam 🔌


jasonacox:
@jasonacox-sam pleaes address Copilot reviews. Also, please conduct a deep analysis of the recent changes (since last tag) to ensure we are not introducing anything brittle, thread un-safe, or poorly engineered. Additionally, make sure the control instructions in README make sense for the user to be able to add and use.

jasonacox-sam:
Copilot's review is addressed and the deep-dive is done — commit ff595b1 pushed (branch rebased onto current `main` first, full suite: **289 passed**).

## 1. Copilot review — fixed

The Basic LAN row in the mode table implied `PW_EMAIL` alone enables hybrid control. Correct call: the cloud-control connection authenticates with the token files at `PW_AUTH_PATH` (from the one-time `python -m pypowerwall setup`) and has no password to log in with, so email without a token silently leaves control unavailable — monitoring unaffected, `/control/*` returns errors.

- Mode-table row now matches the TEDAPI rows: **add `PW_EMAIL` + `PW_AUTH_PATH`** for hybrid cloud control.
- Added a **"Basic LAN + Cloud Control (Hybrid)"** walkthrough under the Basic LAN section — complete docker example with the auth volume mount and the one-time setup command, so a user can go from read-only monitoring to working control by following it verbatim.

## 2. Deep analysis — everything since v0.5.0

Reviewed all 14 commits on `main` since the tag plus this PR: Basic LAN support (#80), the trend legend/tooltip UI (#81), `PW_TIMESERIES_PATH` directory handling, local control via mapped `set_*` methods (#83), and this hybrid change. **No blocking problems found.** Details:

**Thread safety — sound.**
- All control write paths (`cloud_control()`, `local_control()`, and the raw-`post` fallback) serialize through the *same* `GatewayManager._write_lock`, so a `set_operation()` read-modify-write race can't open up between the cloud and local paths. Writes also run through the dedicated executor with per-call timeouts.
- The cloud-control connection is created in a named background task with a 15s cap, so a slow Tesla login can't block startup or health checks; on failure it degrades cleanly (warning log, control 503s, monitoring untouched). `_cloud_control` is a single attribute assigned/consumed on the event loop — no torn state.
- Poller hardening from #80 is correct: per-step timeouts sit above the library's internal request timeout (comment explains why — avoids a stuck thread holding pypowerwall's per-function API lock), and the "aggregates returned None/ERROR = real failure" check prevents a dead gateway from being marked online with empty data.

**Hybrid correctness — verified against pypowerwall 0.17.0.**
- The lazy-init local connection passes `email`/`authpath` alongside `host` + `password`, but pypowerwall's mode selection goes local whenever `host` is set (no `cloudmode`), and the password-only branch builds the local Basic LAN client — so **reads stay 100% local** in hybrid mode. ✓
- The new gating condition mirrors exactly the credential OR-logic (`gw_pwd` / `rsa_key_path` / `password` / global `PW_PASSWORD`) used for both config validation and the actual connection build — no drift between "qualifies for hybrid" and "connects locally".
- Bonus fix included: v1r configs with an RSA key but no `gw_pwd` previously never got the cloud-control connection (old gate required `gw_pwd`); they do now.

**One nuance worth knowing (pre-existing, not introduced here):** pypowerwall's `connect()` falls back local → fleetapi → cloud, so if the local link drops *and* token files are present, the monitoring connection can shift to cloud reads until local recovers. That's also the resilience path for the existing v1r hybrid; noting it for the record, no change made.

**Brittleness sweep — clean.**
- UI legend/tooltip listeners attach once to static DOM (no accumulation across re-renders); keys are real `<button>`s; the no-data path clears the tooltip.
- Timeseries directory-path resolution appends `timeseries.db` and passes `:memory:` through — covered by tests.
- Basic LAN endpoint skips are keyed off `gateway.basic_lan` set at registration, so non-Basic-LAN gateways keep full polling (regression-tested).

## 3. README control instructions — verified

Walked the docs as a user would: the mode table, the per-mode examples, the "Cloud Authentication Setup" section, and the control-security notes (`PW_CONTROL_SECRET`) are consistent and each mode's example is copy-pasteable. The one gap was exactly what Copilot flagged — Basic LAN had no control walkthrough at all; it now does. Hybrid is documented consistently across the table, the new subsection, and RELEASE.md (which already said "with `PW_AUTH_PATH`/token as usual").

Ready for another look — happy to fold in any feedback. PW3 Basic LAN hybrid still wants a hardware tester if anyone's rig is set up for it.

— Sam 🔌

nesys:
looking into this

jasonacox-sam:
Thanks nesys — appreciate you picking up the hardware test, especially since this whole hybrid path came out of your discussion #79. 

Since this is still only on the PR branch (not in the published image yet), the fastest way to run it:

```bash
git clone https://github.com/jasonacox/pypowerwall-server.git
cd pypowerwall-server
gh pr checkout 85        # or: git fetch origin pull/85/head && git checkout FETCH_HEAD
pip install -r requirements.txt
PW_HOST=<pw3-ip> PW_PASSWORD=<customer-pwd> PW_EMAIL=<your@tesla-account> PW_AUTH_PATH=~/.pypowerwall ./run.sh
```

(or build the docker image from the branch and run it with the env/volume layout in the new **"Basic LAN + Cloud Control (Hybrid)"** README section)

Three things worth verifying on a real rig:

- **Reads stay local** — monitoring keeps flowing even with WAN unplugged (Basic LAN reads need no internet)
- **Control goes through the cloud** — `POST /control/reserve` (and `/control/mode`, `/control/grid_charging`) changes take effect; set `PW_CONTROL_SECRET` so writes aren't 403'd
- **Clean degradation** — with cloud auth missing/broken, monitoring is unaffected and control requests return a clear error rather than hanging

If anything behaves oddly, logs + what mode the console shows for the gateway card will tell us where to look. Happy to iterate.

— Sam 🔌

nesys:
I tested PR #85 on my PW3 using Basic LAN for monitoring and cloud for control.

The hybrid architecture itself is working well: Basic LAN telemetry continued streaming normally in the Console while control operations were routed through the Tesla Owner API.

During testing I found two issues/edge cases that may be worth reviewing.

### 2a. `/control/mode` without a companion `level` does not reliably persist the mode change

Starting state:

```
Reserve: 0
Mode: autonomous
```

I called:

```
POST /control/mode
{"value":"self_consumption"}
```

The request returned HTTP 200 and Tesla reported:

```
"real_mode":"self_consumption"
"result":"Operation mode updated"
```

However, the change did not persist. A subsequent cloud read still reported:

`Mode: autonomous`

and the Tesla app remained in Savings.

I also noticed that this standalone `set_mode()` operation generated both a `/backup` and an `/operation` Owner API call, and the response contained:

`"backup_reserve_percent": false`

Using the companion `level` parameter instead:

```
POST /control/mode
{"value":"self_consumption","level":0}
```

caused the server to use `set_operation(0, "self_consumption")`.

This time the change DID persist:

```
Reserve: 0
Mode: self_consumption
```

and the Tesla app immediately changed from Savings to Self-Powered.

So on my PW3, the combined `set_operation(level, mode)` path appears to be more reliable than calling `set_mode()` alone.


### 2b. A reserve/level of 0 appears to be treated as boolean `false`, and can prevent a mode change from persisting

After successfully switching to Self-Powered, I tried to restore the original state with:

```
POST /control/mode
{"value":"autonomous","level":0}
```

Again, the request returned HTTP 200 and:

```
"real_mode":"autonomous"
"result":"Operation mode updated"
```

but the change did NOT persist.

Cloud readback remained:

```
Reserve: 0
Mode: self_consumption
```

and the Tesla app remained in Self-Powered.

The response again showed:

`"backup_reserve_percent": false`

I then repeated the exact same operation using a non-zero reserve:

```
POST /control/mode
{"value":"autonomous","level":1}
```

This worked correctly and persisted:

```
Reserve: 1.0
Mode: autonomous
```

The Tesla app also immediately returned to Savings.

I then set the reserve back to 0 separately, which also worked correctly, leaving the system back in its original state:

```
Reserve: 0
Mode: autonomous
```

This makes me suspect that a reserve/level value of `0` may be converted or interpreted as boolean `False` somewhere in the control path. The API reports success, but in some cases the requested mode does not actually persist.

Summary:

```

> - Hybrid Basic LAN monitoring + cloud control works on my PW3.
> - Basic LAN telemetry remains active while cloud writes are performed.
> - `set_operation(level, mode)` successfully changes the operating mode and the change is visible in the Tesla app.
> - Standalone `set_mode()` did not persist the requested change in my test.
> - `level=0` appears to be an edge case: it is returned as `backup_reserve_percent:false` and can result in a mode change reporting success without actually persisting.
> - Using `level=1` avoided the issue.

```

Happy to run additional tests if useful.

nesys:
Quick follow-up on the remaining hardware tests you requested.

### 1. Reads stay local with WAN unavailable — PASS

I temporarily blocked all WAN traffic from the PR85 container while keeping access to the local/private networks and the PW3 at 10.42.1.44.

Confirmed that the container could no longer reach Tesla:

`curl: (7) Failed to connect to owner-api.teslamotors.com port 443`

At the same time, /pw/aggregates continued returning live data normally from the Powerwall.

After removing the temporary firewall rules, WAN connectivity was immediately restored (owner-api.teslamotors.com reachable again).

So Basic LAN monitoring is confirmed to continue working with no Internet/cloud connectivity.

### 3. Clean degradation without cloud auth — PASS

I also started a separate PR85 container with PW_HOST + PW_PASSWORD but without PW_EMAIL/PW_AUTH_PATH.

Basic LAN initialized normally and /pw/aggregates continued returning live data.

A control request:

```
POST /control/mode
{"value":"self_consumption","level":1}
```

returned immediately with:

```
HTTP/1.1 503 Service Unavailable
{"detail":"Control operation failed or gateway not available"}
```

No hang and no impact on local monitoring.

So from my PW3 hardware testing, both remaining scenarios behave as expected.

Together with the cloud-control testing from my previous comment, the hybrid architecture is working very well here: local Basic LAN for telemetry, cloud only when a write operation is requested.

Happy to run any additional tests if useful.

jasonacox-sam:
Thank you nesys — this is a genuinely excellent test report. The hybrid architecture itself passing (local reads with WAN blocked, clean 503 degradation without cloud auth) is exactly what we hoped for, and your two edge cases traced all the way down to a real root cause. Findings below.

## Root cause of 2a / 2b — confirmed

The server side is clean: `/control/mode` validates `level` with an explicit `isinstance(int)` check (bools rejected), so `{"value": "autonomous", "level": 0}` correctly reaches the library as `set_operation(0, "autonomous")` — no truthiness bug in this repo.

The bug is in `pypowerwall` 0.17.0's shared `set_operation()`:

```python
payload = {
    'backup_reserve_percent': level if level > 0 else False,   # <-- 0 becomes JSON false
    'real_mode': mode
}
```

A legitimate reserve of `0` is coerced to boolean `false` (a v1-local-API-era idiom meaning "disable reserve"). On the cloud path the backend then splits this into **two separate Tesla Owner API commands** — `BACKUP_RESERVE` followed by `OPERATION_MODE` — and Tesla applies them asynchronously. Your observations map perfectly:

- `level: 1` + mode → both commands land → persists ✓
- `level: 0` + mode → the `BACKUP_RESERVE=0` command races the mode change and the mode can be silently dropped (HTTP 200, `"backup_reserve_percent": false` echo, no persistence)
- standalone `set_mode()` → the library **back-fills the current reserve** to avoid changing it, and when that reserve is 0 the same two-command sequence fires — which is why 2a failed too
- standalone `set_reserve(0)` → the mode command re-writes the *current* mode (no-op), so only the reserve write matters → works ✓

So: not PW3-specific, not hybrid-specific — any cloud control write that changes mode while reserve is 0 can hit it.

## What I pushed

Commit 0880e36 (docs):
- **README now shows how to actually call the control API** — curl examples with the auth header for reserve, mode, and the mode+level companion parameter (these didn't exist before), plus a note documenting the reserve-0 caveat and the reliable pattern you derived empirically: set the mode with the current reserve level, then set reserve 0 in a separate call.
- Genericized the example `PW_HOST` values (`<pw3-ip>` placeholders).

Full suite still green: **289 passed**.

## Where the real fix belongs

The proper fix is in the `pypowerwall` library, not here: the cloud backend should send a single `OPERATION_MODE` command for mode-only writes (no reserve back-fill on the cloud path — partial payloads are already supported), and the shared `set_operation()` should stop coercing an explicit `0` to `false`. That's a change with implications for the v1 local/TEDAPI write paths too, so I'm taking it to the pypowerwall repo as its own PR with tests rather than burying it in this one — happy to link it here once it's up.

Thanks again for running the full matrix — including the WAN-block and no-auth degradation tests. This kind of hardware validation is exactly what the hybrid path needed before merge.

nesys:
One more hybrid-mode issue found during testing.

The PR85 Console currently shows:

```
POWERWALL MODE
Self-Consumption
```

However, the Tesla app shows Savings, and a direct cloud read from the same container confirms:

```
Cloud mode   : autonomous
Cloud reserve: 0
```

The server logs also explicitly state that in Basic LAN mode operation mode/reserve are not available locally:

`"Basic LAN mode serves a limited local API - skipping endpoints not exposed in this mode (... operation mode/reserve ...)"`

So the Console mode value appears to be stale or derived from internal state rather than refreshed from the actual cloud state.

This is reproducible after switching modes through the hybrid control API:
```

> - actual Powerwall/cloud state: autonomous
> - Tesla app: Savings
> - PR85 Console: Self-Consumption

```

It may be worth making the hybrid-mode Console read mode/reserve from the cloud connection, or otherwise marking those fields unavailable rather than showing a stale value.

jasonacox-sam:
Good catch again — confirmed and fixed in commit 3fe642e.

**Root cause:** the Console's mode/reserve values come from the poller's cache. Basic LAN has no local operation-mode endpoint (that's the exact skip the log line you quoted refers to), so the poller was pre-filling `mode` from the last cached value — which, after a control write, was whatever *we* last set, never refreshed from the actual system state. Exactly the stale-display you observed.

**Fix (two parts):**
- **Hybrid mode (cloud control available):** mode and reserve now refresh from the cloud control connection on every poll cycle — same lock/timeout path as control writes, so no new thread-safety surface. The Console should now agree with the Tesla app within one poll interval.
- **Plain Basic LAN (no cloud credentials):** the stale pre-fill is dropped entirely rather than served forever — mode/reserve show as unavailable instead of lying.

Two regression tests added covering both paths; full suite green: **291 passed**.

Worth noting the read path is a *cloud* read now in hybrid mode, so like the control writes it needs the cloud connection up — if cloud auth breaks, mode/reserve degrade to unavailable while local monitoring keeps flowing (consistent with the degradation behavior you already validated).

If you get a chance to re-pull the branch, the repro you described (switch modes via control API, then check Console vs Tesla app) should now agree — would appreciate the confirmation.

— Sam 🔌

nesys:
Confirmed on hardware with commit 3fe642e.

The hybrid Console mode/reserve refresh fix is working correctly.

I tested a complete cycle:

```

> - Initial state: autonomous / reserve 0 → Console shows Time-Based
> - Changed through PR85 to self_consumption / reserve 1
> - Cloud readback confirmed self_consumption / 1.0
> - Console automatically updated to Self-Consumption
> - Restored autonomous / reserve 1
> - Then restored reserve separately to 0
> - Final cloud state confirmed autonomous / 0 → Console back to Time-Based

```

So the Console now correctly tracks the actual cloud state in hybrid Basic LAN + cloud control mode. Fix confirmed.

One minor logging observation: the hybrid poll loop currently logs:

```
cloud_control(get_mode) completed successfully
cloud_control(get_reserve) completed successfully
```

at INFO level roughly every 5 seconds.

The polling itself makes sense, and from the logs pypowerwall appears to cache the actual cloud read (the real /site_info HTTP request is much less frequent), but these INFO messages make the log quite noisy.

It might be worth moving successful get_mode/get_reserve polling messages to DEBUG, or only logging when the returned value changes / an actual cloud request occurs.

Thanks — 3fe642e looks good on my PW3.

jasonacox-sam:
Thanks nesys — that full-cycle confirmation of 3fe642e is exactly the verification this needed, and it closes out the stale-Console issue.

Your logging observation was spot on, and it's fixed in 1bf9809:

- **Reads log at DEBUG now** — `cloud_control(get_mode/get_reserve)` succeeds twice per poll cycle, so those lines are gone from INFO. `local_control()` reads get the same treatment for symmetry.
- **Writes stay at INFO** — `set_reserve`, `set_mode`, etc. are user-initiated and rare, so those confirmations remain visible. One INFO line per control operation, zero from polling.

Regression test added covering both sides (poll reads absent at INFO, `set_reserve` still present); full suite: **292 passed**.

That resolves everything raised in this thread so far — hybrid mode confirmed on real PW3 hardware end to end, with all three of your edge cases traced to root causes. The remaining known wart is the reserve-0→`false` coercion in `pypowerwall` itself, which is headed to that repo as its own PR so the fix lands for every consumer; I'll link it here when it's up.

— Sam 🔌

nesys:
Additional hardware validation on 3fe642e.

I tested three states:

### 1) Hybrid mode, cloud healthy
- Basic LAN monitoring works normally
- Cloud mode/reserve refresh works
- Console correctly follows cloud mode changes

### 2) Plain Basic LAN, no cloud credentials
- Backend correctly reports that operation mode/reserve are unavailable:
  "Basic LAN mode serves a limited local API - skipping ... operation mode/reserve ..."
- However, Console still shows:
```
  CONNECT MODE: Basic LAN
  POWERWALL MODE: Self-Consumption
  CONNECTION: Healthy
```

Since mode is not available from Basic LAN and no cloud connection exists, "Self-Consumption" appears to be a UI fallback/default rather than a real value. It would probably be safer to show **N/A / Unavailable**.

### 3) Hybrid mode with WAN intentionally blocked
- Local Basic LAN monitoring continued to work correctly
- Cloud reads failed as expected with connection refused
- Console remained:
```
  CONNECT MODE: Basic LAN
  POWERWALL MODE: Self-Consumption
  CONNECTION: Healthy
```

From a user perspective, this hides the fact that the hybrid connection is degraded.

I would expect something closer to:

```
  CONNECT MODE: Hybrid
  CONNECTION: Degraded
  Local: Healthy
  Cloud: Unavailable
```

For Powerwall Mode / Reserve, either:
- last known cloud value marked as stale, or
- N/A / Unavailable

would be safer than showing a fallback value.

jasonacox-sam:
Both findings confirmed — same root cause, and this time it lives in the presentation layer (the backend data was already honest after 3fe642e).

**Root cause:** `/api/operation` carried a hard-coded `real_mode = "self_consumption"` fallback from the early proxy days. Plain Basic LAN has no local mode endpoint and no cloud link, so the endpoint served that fabricated default — that's the "Self-Consumption" you saw. Same path in the WAN-blocked hybrid case: cloud reads fail, the cache has nothing, the default lied, and the Console rendered it verbatim.

**Fixed in 5fd09c2:**

- `/api/operation` now returns `real_mode: null` when no real source exists. Cached polled mode and `system_status.default_real_mode` remain the fallbacks for gateways that can actually provide them — the fabricated constant is gone. The Console renders `--` (unavailable), covering both your case 2 and the mode part of case 3.
- `/stats` gained a `cloudcontrol` flag (true when the shared cloud-control connection is live), and the Console connect-mode card shows **Hybrid** for Basic LAN + cloud control instead of plain "Basic LAN" — the cloud side is now visible at a glance. The flag is read on the event loop, same single-threaded access pattern the poll loop already uses for `_cloud_control` — no new locking surface.
- README hybrid section updated to match (Hybrid connect mode, `--` on degradation).

Two regression tests added (null-not-fabricated; hybrid stats flag); full suite: **293 passed**.

**On the fuller degraded-state card** (`Local: Healthy / Cloud: Unavailable`, stale-marked values): agreed that's the better end state — but it's a real UI feature (per-link health tracking plus a card layout change), and it deserves its own change with proper tests rather than getting bolted on at the tail of this PR. Tracked in #87 so it doesn't get lost. One note on what you saw: the existing `Connection: Healthy` indicator tracks the *local* poll link, which genuinely was healthy with WAN blocked — the missing piece is the cloud-side row, which is exactly what #87 adds.

If you re-pull, your three states should now read:

- hybrid, cloud healthy → **Hybrid** + real mode from cloud
- plain Basic LAN → **Basic LAN** + `--`
- hybrid, WAN blocked → **Hybrid** + `--`

Thanks for pressing on this one — a status display that fabricates values is worse than one that shows nothing, and this needed real hardware to surface.

— Sam 🔌

nesys:
Additional hardware validation on 5fd09c2.

I re-tested all three states we discussed:

### 1. Hybrid mode, cloud healthy

* Basic LAN monitoring works normally
* Cloud mode/reserve refresh works
* Console shows:
```
    CONNECT MODE: Hybrid
    POWERWALL MODE: Time-Based
    CONNECTION: Healthy
```
* Normal INFO polling noise is gone as expected after 1bf9809.

PASS.

### 2. Plain Basic LAN, no cloud credentials

* Backend correctly reports that operation mode/reserve are unavailable in Basic LAN mode
* Console now shows:
```
    CONNECT MODE: Basic LAN
    POWERWALL MODE: –
    CONNECTION: Healthy
```
* The fabricated Self-Consumption value is gone.

PASS.

### 3. Hybrid mode with WAN intentionally blocked

* Local Basic LAN monitoring continued to work correctly
* Cloud reads failed as expected
* Console changed to:
```
    CONNECT MODE: Hybrid
    POWERWALL MODE: –
    CONNECTION: Healthy
```
* This is consistent with the current semantics: Connection: Healthy represents the local polling link, which genuinely remained healthy, while the unavailable cloud mode is no longer replaced with a fabricated value.

After restoring WAN, the hybrid container recovered automatically without restart and the Console returned to:

```
CONNECT MODE: Hybrid
POWERWALL MODE: Time-Based
CONNECTION: Healthy
```

PASS.

So 5fd09c2 fixes both issues from the previous hardware test: the fabricated mode is gone and Hybrid is now represented correctly.

I have two small observations left from the WAN-failure test:

### 1. Degraded state visibility

Even though:

Hybrid + – + Healthy

is technically correct according to the current semantics, visually it can still look as though the whole hybrid connection is healthy unless the user understands what – means.

Would it make sense to show:

`CONNECT MODE: Hybrid (Degraded)`

whenever Basic LAN is healthy but the cloud side is unavailable?

Connection: Healthy could keep its current meaning as the local connection status. Then #87 could later add the fuller:

```
Local: Healthy
Cloud: Unavailable
```

presentation as planned.

### 2. Cloud failure retry/backoff

The normal INFO polling noise is indeed gone. I only see this issue once WAN/cloud connectivity is lost.

During the outage, get_mode and get_reserve each appear to trigger a failed SITE_CONFIG request on every ~5 second poll cycle. This results in two sets of HTTP fallback / connection-refused warnings and errors roughly every 5 seconds.

Would it make sense to introduce a backoff specifically for cloud reads after a connectivity failure, for example retrying cloud state less frequently while it remains unavailable, while leaving the 5-second local Basic LAN polling completely unchanged?

Overall, the functional degradation/recovery path looks good. The two points above are just UX/failure-mode observations from deliberately taking the cloud side down.

Thanks again — happy to test any follow-up commit on the real PW3.

jasonacox-sam:
Thanks nesys — that three-state validation closes it out. All PASS on 5fd09c2, the fabricated mode is gone, Hybrid is represented honestly, and the degradation/recovery path survives a real WAN outage without a restart. Exactly the verification this needed.

On the two observations — both are real, and I'd handle both, but in #87 rather than here:

**1. Degraded state visibility** — agreed, `Hybrid + – + Healthy` is technically correct but easy to misread as "everything's fine." Rather than add an interim `Hybrid (Degraded)` label in this PR, I'd fold it into #87's per-link display — that gives the actual answer (`Local: Healthy / Cloud: Unavailable`) instead of a second approximation of it. If this PR ends up sitting unmerged for a while, I'm happy to add the interim label as a stopgap.

**2. Cloud failure backoff** — agreed the per-cycle failed SITE_CONFIG chatter every ~5s during an outage is noisy, and a cloud-reads-only backoff (local polling untouched) is the right shape. But that's a polling-behavior change, and this PR is hardware-confirmed as-is — I'd rather not re-open the test cycle for it here. It belongs with #87 or a small follow-up right after this merges.

Keeping this one scoped for merge review: hybrid read/write split, the Console refresh fix, log levels, and the honest mode display.

— Sam 🔌

nesys:
Sounds good to me. I agree with keeping this PR scoped and hardware-confirmed as-is, and handling both per-link degraded-state visibility and cloud-read backoff in #87 / a follow-up.

From my side, 5fd09c2 is fully validated on the PW3 and I don’t see any remaining blocker for merge.

Happy to hardware-test #87 as well when it’s ready

jasonacox-sam:
Thanks nesys — your validation work on this PR has been outstanding, from the initial hybrid test through the WAN-outage and three-state checks. This is exactly the kind of hardware verification that makes the merge case solid.

PR is with Jason for merge review. #87 is up (per-link Local/Cloud health in the Console plus stale-marked mode/reserve values and the cloud-read backoff), and I'll ping you there when it's ready for a test pass.

— Sam 🔌

jasonacox:
Thank you @nesys !! You help and validation is greatly appreciated.

Sam, I'll merget this and move to #87 for review and merge.

Finally, following thread above:

> ## Where the real fix belongs
> The proper fix is in the `pypowerwall` library, not here: the cloud backend should send a single `OPERATION_MODE` command for mode-only writes (no reserve back-fill on the cloud path — partial payloads are already supported), and the shared `set_operation()` should stop coercing an explicit `0` to `false`. That's a change with implications for the v1 local/TEDAPI write paths too, so I'm taking it to the pypowerwall repo as its own PR with tests rather than burying it in this one — happy to link it here once it's up.

@jasonacox-sam  is this still something we need to update? If so, can you please create that PR and update this thread with that link?

jasonacox-sam:
Yes — still needed. I re-checked `main` today: the reserve-0 → `False` coercion and the always-send-both-fields payload construction are unchanged since v0.17.0, so every consumer of `set_mode()` / `set_operation(0, mode)` still carries the race nesys caught on hardware.

The fix is now up: **jasonacox/pypowerwall#377** — https://github.com/jasonacox/pypowerwall/pull/377

What it does:

- **Partial payloads for cloud / fleetapi / tedapi** — `set_operation()` writes only the fields the caller set. A mode-only write now sends a single `OPERATION_MODE` command (no reserve back-fill), a reserve-only write a single `BACKUP_RESERVE` command. That removes the two-racing-commands failure mode behind both of nesys's findings (2a and 2b), and as a bonus cuts the duplicate Tesla audit-log entries.
- **Explicit `level=0` stays numeric `0`** — the `level if level > 0 else False` coercion is gone. This also fixes `set_reserve(0)` silently no-opping on the local v1 path, where `False` means "leave unchanged".
- **Local gateway path keeps full-overwrite semantics** — omitted fields still get back-filled with current values (raw scale), so PW2 local behavior is unchanged.
- **Cloud + FleetAPI validation hardening** — the truthiness-based "missing parameters" guard rejected a legitimate `{'backup_reserve_percent': 0}` partial payload; it now checks `is None`.
- Full library suite: **404 passed, 10 skipped**, with new regression tests for each of the above.

Once #377 merges and ships in a release, pypowerwall-server's `/control/mode` will work standalone (no companion `level` needed) even at reserve 0 — the README caveat we documented in this PR can then be dropped in the follow-up that bumps the dependency.

Thanks again @nesys — the hardware traces that isolated this (the `backup_reserve_percent: false` echo, the level=1 vs level=0 split) are what made the root cause provable from the library side.

— Sam 🔌

jasonacox:
Merged! 

@jasonacox-sam  and @nesys  - Thanks for the help with this! Good addition. Happy to get this added.  We will fold this into release v0.6.0 with https://github.com/jasonacox/pypowerwall-server/pull/88

jasonacox:
@jasonacox-sam  Please create new tag and github release for this new version v0.6.0 - include changes since last tag and contributors. Post on X.

* Container:  jasonacox/pypowerwall-server:0.6.0
* PyPI: https://pypi.org/project/pypowerwall-server/0.6.0/

jasonacox-sam:
**v0.6.0 is out** — tag and release live: https://github.com/jasonacox/pypowerwall-server/releases/tag/v0.6.0

- Release notes cover everything since v0.5.0: Basic LAN support (#80), hybrid mode (#85), per-link hybrid health (#88), the trend legend toggle (#81), optional rate limiting (#84), the v1r local-control fix (#83), and the Console honesty fixes from the hardware test cycle — with contributors credited (@nesys's PW3 validation work, @anupamme's rate limiter, @thcousins's v1r report and hardware verification).
- Container `jasonacox/pypowerwall-server:0.6.0` and [PyPI 0.6.0](https://pypi.org/project/pypowerwall-server/0.6.0/) are both live.
- The library-side fix from this thread is up for review in jasonacox/pypowerwall#377 — once it merges and ships in a pypowerwall release, a follow-up here bumps the dependency and drops the README reserve-0 caveat.
- Release announcement posted to X.

Thanks for the merge — glad to see the #79 design land as a release end to end.

— Sam 🔌

PR #94, opened 2026-09-05
Add Powerwall Control card

erikgieseler:
### Summary
Adds Reserve + Mode control (`TOKEN=$PW_CONTROL_SECRET` curls from the README) to the web Console. The card is only shown when Control Mode is active and drives the existing `POST /control/*` API.

> **Note:** this PR uses the legacy control API (`POST /control/*`).

### Changes
- Backend — `app/api/legacy.py`: new unauthenticated `GET /control/status` → `{"enabled": bool}` (from `settings.control_enabled`). Exposes no secret; all writes still require `verify_control_token`.
- Frontend — `app/static/index.html` (additive only): new `Powerwall Control` card after System Health, hidden until `enabled == true`. Mode select (Self-Consumption / Backup / Time-Based), reserve slider + number (0–100), token field (`sessionStorage` only, eye/eye-off toggle, clear button, `Authorization: Bearer <token>` per request). One Save button: both changed → single `POST /control/mode {"value", "level"}`; reserve 0 + mode change → auto-split into 2 calls; otherwise single call. Values from `GET /api/operation` (incl. stale handling), warning banners, inline status, 30 s refresh with dirty-check, 401 clears the token. Controls the default gateway.
- Tests — `tests/test_api_legacy.py`: 3 new status-endpoint tests.
- Docs — `README.md`: short Web Console paragraph.

### Testing
- `pytest`: 317 passed. Console JS: `node --check` clean.
- Manual: without `PW_CONTROL_SECRET` no card; with secret set reserve/mode; wrong token → 401.

Maintainers are welcome to modify this PR as needed.

Screenshot 2026-09-05 192818

### DIFF
--- a/README.md
+++ b/README.md
@@ -664,6 +664,17 @@ curl -X POST http://localhost:8675/control/mode \
 > stick). If you need mode + reserve 0, set the mode with the *current* reserve
 > level first, then set the reserve to `0` in a separate call.
 
+**Web Console (`/console`):** when `PW_CONTROL_SECRET` is set, the Console shows
+a *Powerwall Control* card (after System Health) with mode select
+(Self-Consumption/Backup/Time-Based), reserve slider + number (0–100) and a
+token field (`sessionStorage` only, `Authorization: Bearer <token>` per request).
+Availability is checked via unauthenticated `GET /control/status`
+(`{"enabled": bool}`); current values come from `GET /api/operation`. One Save
+button sends a single combined `POST /control/mode {"value": mode, "level":
+reserve}` when both changed (reserve 0 + mode change is auto-split into two
+calls, see note above), otherwise a single `/control/reserve` or `/control/mode`
+call. Controls the default gateway.
+
 ### Data Aggregation Strategy
 Multi-gateway aggregation uses **smart aggregation** that will evolve over time:
 
--- a/README.md
+++ b/README.md
@@ -667,7 +667,9 @@ curl -X POST http://localhost:8675/control/mode \
 **Web Console (`/console`):** when `PW_CONTROL_SECRET` is set, the Console shows
 a *Powerwall Control* card (after System Health) with mode select
 (Self-Consumption/Backup/Time-Based), reserve slider + number (0–100) and a
-token field (`sessionStorage` only, `Authorization: Bearer <token>` per request).
+token field (kept in the tab by default, optional “Remember my token on this
+device” persists it in `localStorage`; sent as `Authorization: Bearer <token>`
+per request).
 Availability is checked via unauthenticated `GET /control/status`
 (`{"enabled": bool}`); current values come from `GET /api/operation`. One Save
 button sends a single combined `POST /control/mode {"value": mode, "level":
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,6 +2,11 @@
 
 ## Version History
 
+### [0.6.2] - Upcoming
+
+**Added:**
+- **Web Console Powerwall Control card (`PW_CONTROL_SECRET`)** — the Console gains a Powerwall Control card (shown after System Health when a control secret is configured) with mode select (Self-Consumption/Backup/Time-Based), reserve slider + number (0–100), and a token field kept client-side (session-only by default, optional "Remember my token on this device" via `localStorage`). Mode + reserve changes go through the existing `POST /control/*` API with `Authorization: Bearer <token>`; a mode change together with reserve 0 is auto-split into two calls to avoid the Tesla cloud API quirk that can silently drop the mode change. Card availability is exposed by a new unauthenticated `GET /control/status` returning only `{"enabled": bool}`. (#94)
+
 ### [0.6.1] - 2026-09-04
 
 **Added:**
--- a/app/api/legacy.py
+++ b/app/api/legacy.py
@@ -2183,3 +2183,4 @@ async def get_version():
         pass
 
     return {"version": version, "vint": vint}
+
--- a/app/api/legacy.py
+++ b/app/api/legacy.py
@@ -24,7 +24,8 @@ Key Routes (all cache-backed for graceful degradation):
 Auth Routes (powerflow web app compatibility):
     - POST /api/login/Basic -> Fake login; sets long-lived AuthCookie/UserRecord
 
-Control Routes (require authentication):
+Control Routes (require authentication, except status):
+    - GET /control/status -> Control availability (unauthenticated, {"enabled": bool})
     - POST /control/{path} -> Control operations (reserve, mode, etc.)
 
 Design Principles:
@@ -63,6 +64,20 @@ from app.utils.stats_tracker import stats_tracker
 logger = logging.getLogger(__name__)
 
 router = APIRouter()
+
+
+@router.get("/control/status")
+async def control_status():
+    """Control availability for the Console WebGUI (unauthenticated).
+
+    Returns only ``{"enabled": bool}`` derived from
+    ``settings.control_enabled`` (i.e. ``PW_CONTROL_SECRET`` is set).
+    Deliberately contains no secret and no gateway details because
+    ``/console`` and ``/stats`` are unauthenticated — the browser needs
+    to know *whether* to show the Control card without learning the token.
+    All actual writes stay behind ``verify_control_token``.
+    """
+    return {"enabled": settings.control_enabled}
 
 
 @router.post("/control/{path:path}")
@@ -2168,4 +2183,3 @@ async def get_version():
         pass
 
     return {"version": version, "vint": vint}
-
--- a/app/static/index.html
+++ b/app/static/index.html
@@ -730,7 +730,7 @@
             display: grid;
             grid-template-columns: repeat(auto-fit, minmax(240px, 1fr));
             gap: 16px;
-            margin-bottom: 16px;
+            margin-bottom: 0;
         }
         .control-field {
             background: var(--bg-secondary);
@@ -739,6 +739,12 @@
             display: flex;
             flex-direction: column;
             gap: 10px;
+        }
+        .control-save-row {
+            display: flex;
+            align-items: center;
+            gap: 12px;
+            margin-top: 16px;
         }
         .control-label {
             font-size: 0.75em;
@@ -787,6 +793,18 @@
         }
         .control-token-row input {
             flex: 1;
+        }
+        .control-remember {
+            display: flex;
+            align-items: center;
+            gap: 8px;
+            font-size: 0.8em;
+            color: var(--text-secondary);
+            cursor: pointer;
+            user-select: none;
+        }
+        .control-remember input {
+            accent-color: var(--accent-green);
         }
         .control-btn {
             background: transparent;
@@ -842,8 +860,8 @@
         }
         .control-message {
             font-size: 0.85em;
-            min-height: 1.4em;
-            margin-top: 12px;
+            min-height: 1em;
+            margin-top: 8px;
             color: var(--text-secondary);
         }
         .control-message.success { color: var(--accent-green); }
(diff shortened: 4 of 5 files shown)


jasonacox-sam (approved):
Reviewed and tested locally: checked out the branch, full suite passes (317/317), matching what you reported.

**What I like:**
- `GET /control/status` returning only `{"enabled": bool}` is the right call — unauthenticated, but leaks nothing beyond "control is configured", and the docstring shows you thought about that tradeoff explicitly. Good.
- Reserve-0 + mode-change auto-split mirrors the Tesla API quirk documented in the README, and the inline warning explains *why* the two calls happen. Nice touch.
- Stale-value handling and dirty-check on the 30s refresh so user input isn't clobbered — well done.
- 401 clearing the stored token is correct hygiene.
- Tests cover enabled/disabled/no-secret-leak paths. Solid.

**Minor, non-blocking:**
- In `saveControl()`, `modeChanged`, `reserveChanged`, and `bothChanged` are computed and then discarded (`void` statements) — the actual logic re-derives `mChanged`/`rChanged` a few lines below. Dead code; worth cleaning up in a follow-up but nothing functional.

Happy to see Console control land on the legacy API. @jasonacox — this looks merge-ready from my side; the new unauthenticated status endpoint is the one design decision you may want to eyeball (I think it's fine — no secret, no gateway details exposed).

— Sam 🔌

erikgieseler:
### Added
- `Remember my token on this device` checkbox: token stays session-only by default, persists in `localStorage` when enabled (choice itself is remembered; toggling moves an entered token; clear and 401 wipe both stores).
- Save button now only appears when there are unsaved changes (hides again after save/reload).

### Changed
- Intro line under the card header removed (incl. the `Default: Default Gateway` label logic).
- Wording simplified to plain English: `Mode` / `Token`, `Now:` instead of `Current:`, `Control enabled` badge, `Enter your control token first.`, `Reserve must be 0–100.`, `Nothing to save.`, `Saved. Takes a few seconds to apply.`, `Wrong token.`, `Control is off — …`, `Please check your input.`, `Powerwall unreachable.`, `Tip: reserve 0 plus a new mode is saved in two steps, so nothing gets lost.`, `Might be outdated — …`.
- Token placeholder now `Paste your control token`; technical header hint replaced by the remember checkbox.
- Reserve inputs gained `aria-label`s; `(%)` dropped from the label.
- Spacing aligned with the System Health card (grid margin moved to the save row; status line compact and hidden when empty).
- Removed dead change-detection variables in `saveControl()` (maintainer note); single `mChanged`/`rChanged` pair remains.
- README: token paragraph updated for the remember option.

### Unchanged
- Backend (`GET /control/status`, `POST /control/*`) and tests: no functional changes.

### Verified
- `pytest`: 317 passed. Console JS: `node --check` clean.

jasonacox-sam:
Both Copilot flags look already addressed in the latest push (aacacf5):

- **Reserve-0 split:** the current `saveControl()` handles `mChanged && rChanged && reserveVal === 0` as two calls — mode first using the previous reserve if known, mode-only if `controlInitial.reserve` is null, then `/control/reserve {"value": 0}`. That's exactly the split Copilot suggested; the combined `{value, level: reserveVal}` write now only happens when `reserveVal > 0`.
- **README backticks:** the `Authorization: Bearer <token>` span is balanced in the current head.

So no changes needed — the review was against an earlier commit. Still holding for Jason's merge call.

— Sam 🔌

jasonacox (changes requested):
## Full review (verified locally at head `aacacf5`)

Thanks @erikgieseler — the feature design is good and honors this repo's architecture, but the current head is **broken** and must not merge as-is.

### 🔴 Blocking: JavaScript syntax error — the entire Console is dead

The merge commit `aacacf5` mangled `saveControl()` in `app/static/index.html`. At the end of the `else if (mChanged && rChanged)` block, the brace closes the chain and an orphaned reserve-only block follows with **no `} else if (rChanged) {` header**:

```js
                }            // ← closes else-if chain
                    const r = await postControl('reserve', { value: reserveVal }, token);   // ← orphaned
                    ...
                } else {     // ← stray brace → SyntaxError
```

I extracted the inline script from the branch head and ran `node --check`:

```
SyntaxError: Missing catch or finally after try
```

Because this is the page's single inline script, the error kills **all** Console JS — every panel (status, health, trend, MQTT), not just the new card. The "node --check clean" statements in the PR description and review were evidently against an earlier commit; the `7aa84fb` refactor + merge reintroduced/duplicated code. Note the 317 passing backend tests can't catch this — CI green is misleading here.

**Also part of the same botched merge:** the second branch `else if (mChanged && rChanged)` contains an inner `if (reserveVal === 0)` two-step path that is **unreachable** (the first branch `mChanged && rChanged && reserveVal === 0` already catches it) — duplicated leftover code.

**Fix:** restore the intended chain and delete the duplicate:

```js
if (mChanged && rChanged && reserveVal === 0) {
    // two-step: mode (with previous reserve if known), then reserve 0
} else if (mChanged && rChanged) {
    // single combined POST /control/mode {value, level}
} else if (rChanged) {
    // POST /control/reserve {value}
} else {
    // POST /control/mode {value}
}
```

Please re-verify with `node --check` against the extracted script **at the final head**, and load `/console` in a browser once with and once without `PW_CONTROL_SECRET`.

### 🟡 Required: release note

Per repo policy, add an entry to `RELEASE.md` under `### [0.6.0] - Upcoming` → **Added** describing the Console Control card (no version bump — that happens at release time).

### ✅ What's good (once fixed, this is merge-worthy)

- **Architecture honored**: `GET /control/status` reads `settings.control_enabled` only — no blocking pypowerwall calls in the request path; current values come from the cached `/api/operation`; all writes stay behind `verify_control_token`. 
- **Performance**: the only recurring cost is one extra `/api/operation` fetch per 30 s refresh — a cache read, negligible. `/control/status` is fetched once at load.
- **Security tradeoffs are sensible and documented**: the unauthenticated status endpoint leaks only a boolean; token is client-side only (`sessionStorage` default, opt-in `localStorage`), never rendered, cleared on 401.
- Reserve-0/mode-change auto-split matches the documented Tesla API quirk; dirty-check prevents the refresh from clobbering user input; stale-value banner reuses the `/api/operation` stale contract from 0.6.0.
- Backend tests are well-targeted (enabled / disabled / no-secret-leak). Full suite passes locally (317).

@jasonacox-sam — worth re-checking out the head; the approval was against a pre-merge state of the branch.

jasonacox-sam:
Confirming the blocking review above against the head — **confirmed, and the fix is already pushed** as [853926b](https://github.com/erikgieseler/pypowerwall-server/commit/853926bce419cf7ddfd83a47a6c0476fe7518a1d).

Re-checking `aacacf5` first-hand: the merge dropped the `} else if (rChanged) {` header, so an orphaned reserve-only block follows the closed chain — the page's single inline script fails `node --check` with `SyntaxError: Missing catch or finally after try`, which kills **every** Console panel, not just the new card. My earlier approval and my "no changes needed" note above were both made against pre-merge content while the branch was moving underneath them — that one's on me. Re-verified properly this time at the new head.

What the fix commit does (branch allows maintainer edits, so this went straight to the PR branch):

- Restored the intended four-branch chain: **reserve 0 + mode change → two-step** (mode with the previous reserve when known, mode-only when not, then `/control/reserve {"value": 0}`); both changed with reserve > 0 → single combined write; reserve-only and mode-only → their single calls.
- Removed the unreachable duplicate `if (reserveVal === 0)` path inside the combined branch — the first branch now guarantees `reserveVal > 0` there.
- Added the release note under `### [0.6.2] - Upcoming` in `RELEASE.md` — 0.6.0 and 0.6.1 are already released, so the upcoming section starts at 0.6.2; no version bump.

Verified at the new head: `node --check` clean on the extracted script, full suite **317/317 passing**.

On the Copilot review, both points check out against the earlier commit and both are resolved now: the reserve-0 bypass was real, and the restored first branch is exactly the always-split-when-zero logic Copilot suggested; the README `Authorization: Bearer <token>` backtick imbalance is already balanced at the final head — no change needed there.

@erikgieseler — none of the breakage was your design's fault. The brace mangling came in through the merge commit, not your refactor, and the design itself is genuinely good: the auto-split for the reserve-0 quirk, stale-value handling, 401 token hygiene, and enabled-only gating are all exactly right, and the tests cover the meaningful paths. If you get a chance, loading `/console` once with and once without `PW_CONTROL_SECRET` on the new head would be a good final real-browser confirmation — but from the code side this now matches what you intended.

— Sam 🔌

jasonacox (approved):
Re-verified at the new head `853926b` — all blocking items resolved. **Approved.** ✅

- Extracted the inline Console script and ran `node --check`: **clean**.
- The restored four-branch chain in `saveControl()` is exactly the intended logic (reserve-0 + mode → two-step; both changed with reserve > 0 → single combined write; reserve-only; mode-only), and the unreachable duplicate reserve-0 path is gone.
- Release note added under `### [0.6.2] - Upcoming` with the `(#94)` reference — correct placement and no version bump. 
- Full suite: 317/317 passing.

@jasonacox-sam — nice recovery. Owning the miss directly, re-verifying first-hand instead of defending the earlier approval, shipping a surgical fix with a clear commit message, *and* getting the release note right — that's exactly how a review miss should be handled. The lesson here isn't really about the miss itself: reviewing a moving branch is how everyone gets burned eventually. It's that verification claims should always be pinned to a commit SHA.

One follow-up I'd like you to take on: this whole class of bug was catchable by machine, not by eyes. **Please add a CI step that syntax-checks the Console's inline JavaScript.** Node is preinstalled on `ubuntu-latest`, so it can be a cheap job in `pytest.yml` (or a tiny separate workflow) — extract inline `<script>` blocks from `app/static/index.html` and run `node --check` on each:

```yaml
  js-syntax:
    runs-on: ubuntu-latest
    timeout-minutes: 5
    steps:
      - uses: actions/checkout@v4
      - name: Check inline JS syntax in static HTML
        run: |
          python3 - <<'EOF'
          import re, subprocess, sys, glob
          failed = False
          for path in glob.glob('app/static/**/*.html', recursive=True):
              html = open(path).read()
              for i, s in enumerate(re.findall(r'<script(?![^>]*src)[^>]*>(.*?)</script>', html, re.S)):
                  tmp = f'/tmp/{i}.js'
                  open(tmp, 'w').write(s)
                  r = subprocess.run(['node', '--check', tmp], capture_output=True, text=True)
                  if r.returncode != 0:
                      print(f'{path} inline script #{i}: SYNTAX ERROR\n{r.stderr}')
                      failed = True
          sys.exit(1 if failed else 0)
          EOF
```

Note `pytest.yml` currently has `paths-ignore` for `**/*.md` only, so HTML changes already trigger it — a new job there slots in cleanly. Happy to review the workflow PR when you have it.

@erikgieseler — thanks for the contribution and for allowing maintainer edits; this is a genuinely useful addition to the Console. Good to merge from my side.

jasonacox-sam:
CI guard is in: **[PR #95](https://github.com/jasonacox/pypowerwall-server/pull/95)** adds the `js-syntax` job to `pytest.yml` — extracts every inline `<script>` block from `app/static/**/*.html` and runs `node --check` on each. ~30s, no dependencies beyond preinstalled Node, and HTML changes already trigger this workflow since `paths-ignore` only covers markdown and `docs/`.

Took the snippet from the review nearly verbatim — the only hardening is `sorted(glob)` for deterministic run order and per-file tmp names so blocks from different files can't collide. Verified locally by running the exact embedded step from the workflow: all 6 inline scripts across the 4 static HTML files (`index.html`, `example.html`, `powerflow/index.html`, `powerflow/example.html`) parse clean — including this PR's head at `853926b` — and a deliberately broken block fails with the file, the block index, and the actual `SyntaxError`, which is what a red run should show.

Thank you for the thorough re-review and for pinning every verification claim to a SHA — that discipline is the real lesson here, and the guard is the honest consequence of it: this class of bug was always a machine problem, and now the machine catches it instead of the reviewers' eyes.

@erikgieseler — once #95 merges, subsequent workflow runs on this PR will include the new job; your current head already passes it (verified locally, no changes needed).

— Sam 🔌

erikgieseler:
Verified in my browser with the latest fixes, live on my 2 Powerwall 2 units:

- Without `PW_CONTROL_SECRET`: no Control card, rest of the Console unaffected.
- With `PW_CONTROL_SECRET`: card shows, mode + reserve save correctly.

Runs on my 2 Powerwall 2 units without issues.

jasonacox:
Thanks for this @erikgieseler, great addition! 🙏

jasonacox:
@jasonacox-sam This is worth a minor patch release update. Set the version and update release notes. Once that is done I'll create the container.

jasonacox-sam:
Release PR is up: **[#96 — v0.6.2: Web Console Powerwall Control card (#94)](https://github.com/jasonacox/pypowerwall-server/pull/96)** — once it merges, 0.6.2 is ready for the container build.

Per the release procedure in `AGENTS.md`:
- `app/config.py` → `SERVER_VERSION = "0.6.2"`
- `pyproject.toml` → `version = "0.6.2"`
- `RELEASE.md` → `[0.6.2]` section dated **2026-09-05** — the notes themselves were already in place from the fix commit on the PR branch, and #95's CI job is internal-only so it stays out of the notes per the release-note policy.

Verified: 317/317 tests passing at `d5676dd`. The date assumes the container is cut today — one-line adjustment if it slips.

Thanks again to @erikgieseler — browser-verified on live PW2 hardware before merge; this one's ready to ship.

— Sam 🔌

PR #95, opened 2026-09-05
ci: syntax-check inline JS in static HTML

jasonacox-sam:
Follow-up from #94, where a botched merge mangled the Console's single inline `<script>` block (`saveControl()` else-if chain) and `node --check` would have caught it — 317 passing backend tests could not.

## Changes

Adds a `js-syntax` job to `pytest.yml` (snippet from the #94 review, lightly hardened):

- Extracts every inline `<script>` block (skipping `src=` references) from `app/static/**/*.html` and runs `node --check` on each
- Deterministic order (`sorted(glob)`), per-file tmp names so blocks from different files can't collide
- ~30 s, no dependencies beyond preinstalled Node — HTML changes already trigger this workflow since `paths-ignore` only covers `**/*.md` and `docs/**`

## Verification

Ran the exact embedded step locally against the current tree — all 6 inline scripts across 4 HTML files (`index.html`, `example.html`, `powerflow/index.html`, `powerflow/example.html`) parse clean, and the PR #94 head (`853926b`) passes as well. Also negative-tested against a deliberately broken block (fails with file + block index in the message).

— Sam 🔌

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -44,3 +44,26 @@ jobs:
 
       - name: Run pytest
         run: pytest -q
+
+  js-syntax:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Check inline JS syntax in static HTML
+        run: |
+          python3 - <<'EOF'
+          import re, subprocess, sys, glob
+          failed = False
+          for path in sorted(glob.glob('app/static/**/*.html', recursive=True)):
+              html = open(path).read()
+              for i, s in enumerate(re.findall(r'<script(?![^>]*src)[^>]*>(.*?)</script>', html, re.S)):
+                  tmp = f'/tmp/{path.replace("/", "_")}.{i}.js'
+                  open(tmp, 'w').write(s)
+                  r = subprocess.run(['node', '--check', tmp], capture_output=True, text=True)
+                  if r.returncode != 0:
+                      print(f'{path} inline script #{i}: SYNTAX ERROR\n{r.stderr}')
+                      failed = True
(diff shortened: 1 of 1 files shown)


jasonacox (approved):
Verified locally against `main` — **approved**. ✅

- Ran the exact embedded step: 6 inline scripts across the 4 static HTML files, all parse clean — matches your report precisely.
- Negative test: a deliberately broken block (an orphaned `else` after `try`, i.e. the same error class as #94) fails with exit 1 and prints the file, block index, and the real `SyntaxError` — exactly what a red run should show.
- Your two hardenings (`sorted(glob)` for determinism, per-file tmp names to prevent block collisions) are both correct improvements over my snippet.
- Placement in `pytest.yml` is right: HTML already triggers the workflow (`paths-ignore` covers only markdown/docs), it's a parallel job so pytest runtime is unaffected, and 5-min timeout with zero extra dependencies keeps it cheap.

One future-proofing note, **non-blocking**: the regex feeds every non-`src` `<script>` block to `node --check`, which parses as a classic script. If the Console ever gains a `type="application/json"` data block or a `type="module"` script (with `import`/`export`), the job would false-fail. Neither exists today, so nothing to do now — just a known limitation to revisit if it ever bites.

CI-only change, so no `RELEASE.md` entry needed per policy. Nice follow-through, Sam — from review miss to permanent guard in one day is the right trajectory. Merging.

jasonacox-sam:
@jasonacox — CI is green post-merge: the `Pytest` workflow (including the new `js-syntax` job) and the `Simulator` workflow both completed successfully on 2e3c378, and `main` is clean at 28f0577. Thanks for the extra negative test — good to know the failure output names the file and block index. This closes out the #94 follow-up.

— Sam 🔌

PR #98, opened 2026-09-06
feat(ui): attach Powerwall 3 expansion rows to leaders

longzheng:
PW3 expansion packs would show up on the "Powerwall Status" as a separate Powerwall. This PR changes the expansion to appear underneath the leader.

Before

image

After

image

### DIFF
--- a/app/api/legacy.py
+++ b/app/api/legacy.py
@@ -65,6 +65,64 @@ logger = logging.getLogger(__name__)
 router = APIRouter()
 
 
+def _device_serial(value: object) -> Optional[str]:
+    """Extract a serial number from a TEDAPI VIN/DIN or plain serial value."""
+    if not isinstance(value, str) or not value:
+        return None
+    return value.rsplit("--", 1)[-1]
+
+
+def _expansion_parent_indexes(
+    system_status: Optional[dict], tedapi_config: Optional[dict]
+) -> dict[int, int]:
+    """Map battery-expansion block indexes to their leader block indexes.
+
+    TEDAPI config stores expansions under their leader while system status exposes
+    each expansion as a separate ``battery_blocks`` entry.  The mapping lets the
+    legacy flat ``/pod`` response retain the parent relationship for the Console.
+    """
+    if not isinstance(system_status, dict) or not isinstance(tedapi_config, dict):
+        return {}
+
+    blocks = system_status.get("battery_blocks") or []
+    if not isinstance(blocks, list):
+        return {}
+
+    serial_to_index = {}
+    for index, block in enumerate(blocks, 1):
+        if not isinstance(block, dict):
+            continue
+        serial = _device_serial(block.get("PackageSerialNumber"))
+        if serial:
+            serial_to_index[serial] = index
+
+    parent_indexes: dict[int, int] = {}
+    config_blocks = tedapi_config.get("battery_blocks") or []
+    if not isinstance(config_blocks, list):
+        return parent_indexes
+
+    for config_block in config_blocks:
+        if not isinstance(config_block, dict):
+            continue
+        leader_serial = _device_serial(
+            config_block.get("vin") or config_block.get("din")
+        )
+        leader_index = serial_to_index.get(leader_serial)
+        if not leader_index:
+            continue
+        for expansion in config_block.get("battery_expansions") or []:
+            if not isinstance(expansion, dict):
+                continue
+            expansion_serial = _device_serial(
+                expansion.get("din") or expansion.get("vin")
+            )
+            expansion_index = serial_to_index.get(expansion_serial)
+            if expansion_index:
+                parent_indexes[expansion_index] = leader_index
+
+    return parent_indexes
+
+
 @router.post("/control/{path:path}")
 async def control_api(
     path: str, data: dict, authorization: Optional[str] = Header(None)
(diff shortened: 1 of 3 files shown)


jasonacox-sam (approved):
Reviewed — this looks good. Nice UX improvement:

- `_expansion_parent_indexes` cleanly derives the expansion→leader mapping by matching serials between `system_status` and `tedapi_config`, with sensible guards on types and missing serials. Exposing it as `PW{n}_attached_to` keeps the flat `/pod` schema backward-compatible.
- Good call matching on the serial suffix (`rsplit("--", 1)`) rather than full VIN/DIN — the test fixtures show why that matters.
- Frontend re-arrangement keeps unmatched expansions visible at the bottom instead of dropping them, and the `↳` row styling is a clear visual cue.
- Tests cover both the mapped case and the unknown-expansion case; I ran the full `tests/test_api_legacy.py` suite against the branch — 72 passed.

Two nits, neither blocking:
- `app/api/legacy.py` has a stray extra blank line before `pw_status()` and the trailing newline was dropped at EOF.
- `powerwallRowHtml` hardcodes `'Powerwall 3 (Expansion)'` while `pwModel` also has a `BatteryExpansion` branch — future block types would bypass that. Fine for now given PW3-only expansions.

No blocking issues from me. @jasonacox — your call on merge as always.

— Sam 🔌

jasonacox:
@longzheng Thanks! Great addition!

One quiestion, in your screenshots, your after shot is missing the infographics next to the summary data:

image

After:

image

I'm still waiting on a build to test but wanted to know if that was a gap?

longzheng:
> @longzheng Thanks! Great addition!
> 
> 
> 
> One quiestion, in your screenshots, your after shot is missing the infographics next to the summary data:
> 
> 
> 
> image
> 
> 
> 
> After:
> 
> 
> 
> image
> 
> 
> 
> I'm still waiting on a build to test but wanted to know if that was a gap?

I think it's a side effect of the mock. I didn't bother publishing a docker image so I just asked agent to create a mock of what the table will look like.

jasonacox (approved):
## Full review — approved ✅ (and conflicts resolved)

**Merge conflict resolved and pushed** as `7b7ae15` (branch allows maintainer edits). The conflict in `app/api/legacy.py` was a trivial both-sides-add at the same location: this PR's `_device_serial`/`_expansion_parent_indexes` helpers vs. `GET /control/status` from #94 on main — both kept. `index.html` and the tests auto-merged cleanly; I verified the merged Console still contains both the Control card (#94) and the new expansion rows, since both touched the same file. I also added the required `RELEASE.md` entry under `### [0.6.3] - Upcoming` per policy.

**Verified at the merged head:**
- Full suite: **319/319 passing** (317 + the 2 new tests).
- `js-syntax` guard from #95: all inline scripts parse clean — first real exercise of that CI job on a Console change. 🙂
- `tedapi_config` is a real field on `PowerwallData`, and `/api/gateways/` serves it, so the multi-gateway path's `status.data.tedapi_config` lookup works; `/pod` already exposed `PW{n}_PackageSerialNumber` and `PW{n}_Type` on main, so the widened frontend regex has data to match.

**Design-requirements check — all pass:**
- **Cache-first / no blocking calls**: `_expansion_parent_indexes` runs on already-cached `system_status` + `tedapi_config` — pure dict work, no pypowerwall calls in the request path. O(blocks) per `/pod` request with a handful of blocks: negligible, consistent with how `/pod` already assembles its response.
- **No cache mutation**: the backend builds a fresh `pod` dict; the JS multi-gateway path copies blocks via spread (`{ ...block }`) before annotating — the cached objects are never written to. Thread/async-safety is a non-issue: read-only access, no new shared state.
- **Safe defaults**: defensive `isinstance` checks throughout; malformed/missing `tedapi_config` or `battery_blocks` degrades to `{}` → rows render exactly as before (flat list). Non-TEDAPI modes (cloud, Basic LAN) have no `tedapi_config`, so zero behavior change there. Unmatched expansions still render (appended at the end) rather than disappearing — good reliability property.
- **Backward compatibility**: `/pod` only gains `PW{n}_attached_to` keys; nothing removed or renamed.

**On the screenshot question (@jasonacox):** not a gap — the author confirmed the "after" image was an agent-generated mock, and the summary infographics section (battery segment graphics, clock) is untouched by this diff. The merged file retains all 4 `battery-graphic` blocks. Your real-hardware build should look right; that's the one remaining confirmation worth doing since neither of us has a PW3 expansion pack to test against.

**Minor nits, non-blocking:**
- `dict[int, int]` lowercase generics vs. the codebase's `Dict`/`Optional` typing style (fine on Python 3.10+, just inconsistent).
- Expansion rows display the fixed label "Expansion" rather than their `PW{n}` identity — intentional per the screenshot, but worth remembering if anyone ever needs to correlate an expansion row back to `/pod` keys.
- Stray whitespace: extra blank line before `/pw/status`, dropped EOF newline.

Nice contribution @longzheng — the leader/expansion mapping via TEDAPI config serials is exactly the right source of truth, and the frontend refactor (shared `powerwallRowHtml`) actually reduces duplication between the single- and multi-gateway views.

jasonacox:
Local testing:

image

jasonacox:
Thanks @longzheng !

Issue #55, opened 2026-06-23
Concurrent /control/reserve and /control/mode calls silently corrupt gateway config


wabbitro:
**What happens**

When `/control/reserve` and `/control/mode` are called at the same time — which happens naturally when a home automation system fires both as part of a scheduled mode change — the gateway ends up with one field correct and the other wrong. Both calls return HTTP 200. There's no error.

**Why**

Both endpoints route through the same pypowerwall method, `set_operation()`, which always writes `backup_reserve_percent` and `real_mode` together in a single `write_config_v1r` call. When the field it wasn't given is missing, it reads the current value from the 5-second poll cache to fill in the payload.

Under concurrent calls both threads read the same stale cache before either write completes, each builds a full payload with a stale value for the field it didn't own, and whichever write lands last on the gateway wins both fields.

`gateway_manager.py` runs these as blocking calls in a `ThreadPoolExecutor` with no serialization — the code comment even says "All operations use asyncio (no threads/locks needed)", which holds for reads but not writes.

**Reproducing it**

```bash
# Set a known starting state
curl -s -X POST -H "Authorization: Bearer <token>" -H "Content-Type: application/json" \
  -d '{"value": "self_consumption"}' http://<host>:8675/control/mode
curl -s -X POST -H "Authorization: Bearer <token>" -H "Content-Type: application/json" \
  -d '{"value": 0}' http://<host>:8675/control/reserve
# verify: {real_mode: self_consumption, backup_reserve_percent: 0}

# Fire both simultaneously
curl -s -X POST -H "Authorization: Bearer <token>" -H "Content-Type: application/json" \
  -d '{"value": 100}' http://<host>:8675/control/reserve &
curl -s -X POST -H "Authorization: Bearer <token>" -H "Content-Type: application/json" \
  -d '{"value": "autonomous"}' http://<host>:8675/control/mode &
wait
# result: one field right, one field wrong — both returned 200
```

Tested via direct curl and via Home Assistant rest_command service calls (closer to real-world timing). Observed across multiple runs:

```
Run                       | Expected                          | Actual                            | Wrong field
--------------------------|-----------------------------------|-----------------------------------|------------
Real TOU automation       | {reserve=100, mode=autonomous}    | {reserve=0, mode=autonomous}      | reserve
Controlled (curl &)       | {reserve=100, mode=autonomous}    | {reserve=100, mode=self_consumption} + hard error on one call | mode
[...]

PR #104, opened 2026-09-09
Add authenticated HTTP control for PW3 v1r islanding

### LINKED ISSUE #103: Add HTTP islanding control using released pypowerwall methods
Following my islanding contribution in jasonacox/pypowerwall#379, I’d like to prepare a focused PR exposing that capability through pypowerwall-server’s HTTP interface.

My Home Assistant setup already uses a locally added `/control/islanding` endpoint in the legacy proxy, so I have a working reference implementation. I’d adapt that approach to this server’s authentication and request-handling conventions rather than copy it unchanged.

The proposed scope is updating the pypowerwall dependency and adding authenticated islanding/reconnect routing, explicit off-grid confirmation, clear error handling, documentation, and non-actuating tests.

I’m aiming to have a PR ready in the next day or two. Posting here to coordinate and avoid duplicate work—@jasonacox, please let me know if someone is already working on this or if you have a preferred approach.

mcmahonl19:
Closes #103.

Adds `POST /control/islanding` using the released `go_off_grid(confirm=True)` and `reconnect_grid()` methods. Bumps the Docker dependency to pypowerwall 0.17.3 and the package minimum to 0.17.3.

- Uses the existing control token, executor, write lock, and timeout handling.
- Requires `value: "off_grid"` with JSON `confirm: true`, or `value: "on_grid"`.
- Targets the default gateway's local v1r/TEDAPI connection, including in hybrid mode. Existing control routes and cloud behavior are unchanged.
- Returns HTTP errors for invalid input, unavailable backends, failures, and unacknowledged responses; documents that callers must verify grid status and must not automatically retry after a timeout.
- Adds documentation, release notes, and 44 non-actuating test cases. No Console controls, MQTT controls, or Home Assistant configuration changes.

Validation: all 370 backend tests pass on Windows/Python 3.12 with pypowerwall 0.17.3; `pip check` passes and the wheel builds successfully. One existing Starlette/AnyIO deprecation warning remains.

The underlying library implementation was live-hardware validated on a Powerwall 3 using v1r/TEDAPI in jasonacox/pypowerwall#379. This new server HTTP route has only been tested with mocked hardware; no live islanding or reconnect commands were sent while developing this change.

Opening as a draft pending review and hardware validation. @jasonacox-sam, would you be able to validate the HTTP path on your setup?


jasonacox-sam:
Thanks, @mcmahonl19 — this is a careful piece of work, and the honesty up front is appreciated: stating plainly that no live islanding commands were sent during development, alongside 44 non-actuating test cases, is exactly how a contactor-controlling feature should arrive as a draft.

I reviewed the full diff against main (v0.6.4) — the approach looks right:

- **Auth and validation ordering** — `verify_control_token()` runs before any payload parsing, and validation is strict: exact-match `value`, `confirm` must be a real JSON boolean (`"true"`, `1`, and `None` are all rejected), and `off_grid` without `confirm: true` is a 400. The parametrized matrix covers the near-miss payloads, including the `confirmed` typo.
- **Local-only path** — routing through `local_control()` targets the gateway's own v1r/TEDAPI connection even in hybrid mode, with no cloud fallback; the fixture asserts `_cloud_control` and raw `post()` are never touched. Good.
- **Conservative success semantics** — HTTP 200 only on the hardware-observed `result: 1` acknowledgement; anything else is a 502 with the raw library response preserved, and no-response/failure/timeout is a 503. Refusing to present an unacknowledged command as success is the right call for a grid transition.
- **Write serialization** — adding `go_off_grid`/`reconnect_grid` to `_WRITE_METHODS` keeps the #55/#56 lock semantics intact, and the lock test confirms the call waits rather than racing.
- **Route ordering** — the specific route registers ahead of the `/control/{path:path}` catch-all, so it resolves correctly.
- **Docs** — the README warning (ack ≠ confirmation of grid state, verify `grid_status` after the call, never auto-retry after a timeout) is the part most PRs would skip. Dependency bumps are consistent across `requirements.txt` and `pyproject.toml`.

One minor, non-blocking note: the route passes `timeout=10.0` explicitly, which matches `local_control()`'s current default — fine as-is, just worth keeping in sync if the control timeout ever becomes configurable.

On validating the HTTP path: yes — we have a PW3 with a registered v1r key here, so the setup fits. But running contributor code against live hardware is gated on the project owner's explicit approval in this thread, so nothing gets sent to the real Powerwall until Jason signs off here. The underlying library methods were already live-validated in pypowerwall#379, which de-risks the backend — what remains to verify is the HTTP wiring end to end. Until then this stays a draft, which is the right state for it.

— Sam 🔌

mcmahonl19:
Thanks for the thorough review! Glad the approach looks right. Completely understood on needing Jason’s approval for hardware testing—happy to leave this as a draft until then.

jasonacox (approved):
## Full review — code is sound, approved for hardware validation ✅ (verified at head `70676f2`)

For a safety-sensitive endpoint (this one operates the physical grid contactor), this is exactly the shape I want to see in this repo: **simple, strict, and honest about uncertainty**. I verified locally: **370/370 tests pass** (326 + 44 new), the `pypowerwall` methods are confirmed present in 0.17.3 (latest on PyPI), and the branch is current with main.

### Design-requirements check — all pass

- **Not over-engineered**: 48 lines of route + 2 allowlist entries. Everything else rides the existing `local_control()` safety envelope — executor bridge, `asyncio.wait_for` timeout, shared `_write_lock`, never-raises contract. No new machinery invented.
- **Thread/concurrency safety**: `go_off_grid`/`reconnect_grid` added to `_WRITE_METHODS` so islanding serializes with every other control write (and the raw-POST fallback) under the same lock — and there's a dedicated test proving the lock is actually awaited before the executor call.
- **Reliability / safety posture is the standout**:
  - Explicit `confirm: true` (strict JSON boolean — `"true"`/`1` rejected) for off-grid only; reconnect needs none. Right asymmetry.
  - `result: 1` hardware-observed ack required for HTTP 200; anything else is 502 with the raw response in `detail` — no false success.
  - **No fallback paths**: never redirects to the hybrid cloud connection, never falls through to raw POST — both asserted in the fixture teardown (`cloud.mock_calls == []`, `post.assert_not_called()`), which guards every test in the file. Clever pattern.
  - No automatic retry, and the docs explicitly tell callers not to retry after a timeout because the command may still land — the most important sentence in the README addition.
- **Explicit endpoint** (no catch-all), registered before `/control/{path:path}` so it correctly shadows it. Existing control routes untouched.
- **Conventions**: `verify_control_token` reuse, Google-style docstring, `RELEASE.md` entry under `[0.6.5] - Upcoming` with issue reference, dependency bumped consistently in `pyproject.toml` + `requirements.txt`. The 400-vs-422 boundary is documented and tested. Test coverage (14 invalid-payload permutations, auth, disabled, missing-connection/gateway, failure modes, unacknowledged results, lock semantics) is the most thorough of any endpoint in the repo.

### Two small asks — please address before merge

1. **Test fixture consistency**: the fixture sets `gateway_manager._cloud_control = cloud` by direct assignment while using `monkeypatch.setattr` for `control_secret`. To be fair, I checked and this is *not* an actual leak — the autouse `_reset_singleton_state()` in `conftest.py` already resets `_cloud_control` around every test — but please use `monkeypatch.setattr(gateway_manager, "_cloud_control", cloud)` anyway so the fixture doesn't silently depend on conftest internals, and matches the style used two lines above.
2. **Error `detail` shape**: 503 returns a string `detail` while 502 returns a dict (`{message, response}`). If keeping the library response in the 502 payload is the intent (I think it's the right call), please make that explicit with a short comment at the 502 raise, or align the 503 to the same `{message: ...}` shape — either way, so API consumers aren't surprised by the type change between the two errors.

### @jasonacox-sam — go-ahead for the hardware test 🟢

The code side is fully verified; what's left is exactly what mocks can't prove. Suggested sequence on your v1r/TEDAPI PW3 setup:
1. `GET /api/system_status/grid_status` baseline
2. `POST /control/islanding {"value": "off_grid", "confirm": true}` → expect 200 with `result: 1`
3. Confirm grid status flips to island mode within a few poll cycles (and note how long it takes)
4. `POST /control/islanding {"value": "on_grid"}` → expect 200, confirm reconnect
5. One negative probe: wrong token → 401, `confirm: "true"` (string) → 400
6. Watch the server log for the `local_control(go_off_grid) completed` INFO lines and anything unexpected

Please pin your results to the commit SHA you test. Once the two asks above are addressed and hardware validation passes, this is merge-ready.

Thanks @mcmahonl19 — the restraint here (no Console buttons, no MQTT/HA exposure, no retries) is as valuable as the feature itself.

jasonacox-sam:
**Hardware validation complete — all steps pass ✅** (validated at `70676f2`; PW3 + Backup Gateway, firmware 26.26.4, pypowerwall 0.17.3, TEDAPI v1r mode with registered key)

Ran the full suggested sequence on the home system this evening, with the owner's go-ahead earlier in this thread. The house was islanded briefly and reconnected cleanly.

| Step | Result |
|---|---|
| `GET /api/system_status/grid_status` baseline | 200 — `SystemGridConnected` |
| `POST /control/islanding {"value":"off_grid","confirm":true}` | 200 in 181 ms — `{"mode":6,"force":true,"result":1}` |
| Island confirmation | `SystemIslandedActive` on first poll (≤3 s) |
| `POST /control/islanding {"value":"on_grid"}` | 200 in 458 ms — `{"mode":1,"force":false,"result":1}` |
| Reconnect confirmation | back to `SystemGridConnected` after ~21 s |
| Wrong token | 401 `Invalid control token` |
| `confirm: "true"` (string) | 400 `'confirm' must be a boolean` |
| `value: "bogus"` | 400 `'value' must be 'off_grid' or 'on_grid'` |
| `off_grid` without `confirm` | 400 `Off-grid requires 'confirm': true` |

Server log showed exactly the expected lines and nothing unexpected: `[default] local_control(go_off_grid) completed successfully` followed by `[default] local_control(reconnect_grid) completed successfully`.

One real-world observation that proves out the docs: the `on_grid` ack (`result: 1`) came back in ~0.5 s, but the system didn't report `SystemGridConnected` until ~21 s later — contactor reclose is far slower than the acknowledgement. "Acknowledgement is not proof of grid state" isn't boilerplate; it's exactly what a caller will observe. The README warning (verify `grid_status` after the call, never auto-retry after a timeout) is spot-on.

Total off-grid window in the test: ~24 s; final state verified back to `SystemGridConnected`.

Thanks @mcmahonl19 — this is a textbook way to bring a contactor-controlling endpoint to review: 44 non-actuating test cases, no Console/MQTT/HA surface added, honest draft status pending live validation, and docs that anticipated the one behavior the live test actually confirmed (fast ack vs. slow reclose). From my side this is hardware-validated and ready for the owner's merge call.

— Sam 🔌

README.md
```diff
@@ -684,6 +684,36 @@ curl -X POST http://localhost:8675/control/mode \
...
+hardware. Other or missing result codes return HTTP 502 with the library
+response in `detail.response`; unavailable/unsupported connections, exceptions,
+or no response within the existing control timeout return HTTP 503.
+
+**Warning:** these commands operate the physical grid contactor and can interrupt
+power. An acknowledgement is **not** confirmation that the home changed grid
+state. Check `GET /api/system_status/grid_status` after the transition, allowing
+for the configured polling interval. After an error or timeout, the outcome may
+be unknown and the command may still complete; do not automatically retry or
+send the opposite command. No islanding buttons are added to the Console.
```
jasonacox: One more ask: please strengthen this warning to match the wording we just added to the pypowerwall library docs — it spells out the *consequences* (solar dropout, battery depletion while islanded) and the operational rules (no automation, someone present), which this version only implies. Suggested replacement, merging the library warning with your HTTP-specific guidance:

```suggestion
⚠️ **WARNING — Grid island control physically operates your home's grid
contactor. Use with extreme care.** Going off-grid causes a brief transition
(~30s solar dropout) and leaves your home running on battery + solar alone
until you reconnect — if the battery depletes while islanded, your home loses
power. Do not automate these commands without understanding the failure modes;
test only when someone is present. An acknowledgement is **not** confirmation
that the home changed grid state: check `GET /api/system_status/grid_status`
after the transition, allowing for the configured polling interval. After an
error or timeout, the outcome may be unknown and the command may still
complete; do not automatically retry or send the opposite command. No
islanding buttons are added to the Console.
```

jasonacox-sam:
Reviewed the draft — this is clean work. A few notes:

**Code review (positive):**
- Route ordering is right: the `/control/islanding` handler is declared before the `/control/{path:path}` catch-all, so it can't be shadowed.
- Strict input validation follows the `grid_charging` (#29) precedent — 400s on missing/mistyped `value` and `confirm`, and `confirm` must be an actual JSON boolean (no truthy strings). Good.
- `result.get("result") != 1` → 502 rather than reporting success on an unacknowledged response is exactly the right call — the docstring's "acknowledgement is not proof of grid state" caveat and the no-auto-retry guidance on timeout are the correct semantics for an operation with real-world consequences.
- Explicitly targeting `local_control()` rather than the hybrid cloud connection (and asserting zero cloud calls in the tests) avoids the exact class of cloud/local divergence we hit in #85.
- Adding `go_off_grid`/`reconnect_grid` to the locked-methods list in `gateway_manager` keeps the same serialization as the other state-changing calls.
- 44 non-actuating tests covering routing (both hybrid and local), auth, disabled-control, malformed JSON, missing gateways, and failure/timeout paths — that's thorough coverage for a route this consequential.

**On live validation:** happy to run the HTTP path against real hardware, but islanding the site is a live-grid operation, so Jason's approval is the gate before I execute any of this on our network. I've flagged it to him — will report back here once I have a green light.

— Sam 🔌

mcmahonl19:
Jason, here’s proposed wording for the expanded safety warning. It includes Tesla’s documented frequency-shifting behavior near full charge while preserving the HTTP-specific guidance. Does this work for you?

⚠️ WARNING — Grid island control physically operates your home's grid contactor. Use with extreme care. Going off-grid can interrupt power and temporarily interrupt solar production. While islanded, Powerwall may raise the home's electrical frequency to limit or stop solar production, particularly when the battery is near full or available charging capacity is limited. This can cause lights to flicker and affect frequency-sensitive equipment.
Your home depends on available battery and solar power while disconnected from the grid; if those cannot support the load, your home can lose power. Do not assume that a nearly full battery guarantees a seamless transition.
Do not automate these commands without understanding the failure modes; test only when someone is present. An acknowledgement is not confirmation that the home changed grid state: check GET /api/system_status/grid_status after the transition, allowing for the configured polling interval. After an error or timeout, the outcome may be unknown and the command may still complete; do not automatically retry or send the opposite command. No islanding buttons are added to the Console.

mcmahonl19:
Thanks, @jasonacox-sam! One clarification: your earlier comment reports successful hardware validation at `70676f2`, but your later comment says you’re still awaiting approval. Was that later comment just a stale status update? Could you confirm the earlier HTTP hardware-test results remain valid? No additional hardware run requested—just clarifying the record.

jasonacox-sam:
Good catch on the ordering — the later comment was a stale status summary posted out of sequence, and the "awaiting approval" framing in it is superseded: Jason gave the hardware-test go-ahead earlier in the thread, and the validation run at 04:42 UTC was executed *after* that approval.

To confirm the record: the hardware validation results at `70676f2` stand and remain valid — the PR head is unchanged since that run (`70676f2` is still HEAD), so those results apply to the current code as-is. No re-run needed.

— Sam 🔌

jasonacox-sam:
**Record update — and a correction to my own last comment:** the 11:44 reply was accurate when posted, but at 11:58 UTC `f2fd486` ("Address islanding review fixture and error detail comments") landed, so its "the head is unchanged / `70676f2` is still HEAD" line no longer holds. Corrected picture:

- **Approval preceded the test — confirmed.** The hardware-test go-ahead came in the owner's own review at 04:21 UTC (his explicit green light), and the validation run at 04:42 UTC followed it. The later "awaiting approval" comment was a stale, out-of-sequence status note, as covered in the 11:44 reply.
- **Do the `70676f2` hardware results still cover the current head? Yes — on the merits, not by the SHA.** I diffed `70676f2...f2fd486`: the only changes are the `monkeypatch.setattr` fixture fix in `tests/test_islanding.py` and a two-line explanatory comment at the 502 raise in `app/api/legacy.py`. No executable behavior changed, so the validated route — 200 / `result: 1` ack semantics, the 400/401 negative probes, and the off-grid → on-grid cycle — is byte-identical at runtime. No re-run needed on that basis. If a future commit functionally touches the route, I'll re-validate before calling it covered.
- **Both asks from the 04:21 review are now addressed in `f2fd486`:** fixture consistency via `monkeypatch.setattr` for `_cloud_control`, and the 502/503 detail-shape question resolved by documenting the intentional difference in-code at the 502 raise.

The remaining open item is the README safety warning — the expanded wording proposed above is with the owner for his response. That, and the merge call itself, stay with him.

Thanks for pushing on this, @mcmahonl19 — insisting the validation record be exactly straight on a contactor-controlling endpoint is the right instinct, and my own comment needed the correction above.

— Sam 🔌

jasonacox-sam:
Verified the latest Copilot review against head `20fab40` — all three points look valid, with the first being the one that matters most:

1. **Timeout race (real, safety-relevant):** in `local_control()` (`gateway_manager.py:1895`), `asyncio.wait_for()` abandoning the executor future exits the `async with self._write_lock` scope while the underlying thread keeps running. A second request could then acquire the lock and issue the opposite contactor command while the first is still in flight. Tracking the timed-out call as in-flight and rejecting (rather than queueing) subsequent islanding commands until it settles seems like the right shape — for a physical grid-transition operation, rejecting with a "check grid status, retry not advised" error is safer than serializing a queued opposite command behind it.

2. **Boolean ack slips through:** `result.get("result") != 1` accepts `{"result": true}` because `True == 1` in Python. Worth an exact-type check (`isinstance(v, int) and not isinstance(v, bool) and v == 1`) plus a regression test with the boolean response.

3. **README overview:** the mode overview near the top ("All local modes are read-only unless cloud credentials are provided") is now inaccurate given this endpoint — worth calling out the v1r islanding exception there so local-only users don't conclude the endpoint can't work.

— Sam 🔌

jasonacox-sam:
Verified `4cf4bb4a` against the three points from the latest Copilot review — all addressed, and addressed well:

1. **Timeout race** — the `_islanding_futures` tracking plus `asyncio.shield()` is the right shape: the executor future survives `wait_for` cancellation, the `_write_lock` is held until the underlying thread actually completes (released via done-callback), and a second islanding command is rejected with a 409 that tells the caller to check grid status and *not* retry or send the opposite command. I like that other writes (`set_mode` etc.) block until the in-flight call settles rather than being rejected — queueing a benign write is fine; queueing an opposite contactor command is not. The async test covering exactly that distinction (opposite command rejected, queued write deferred then completed) is a nice touch.
2. **Boolean acknowledgement** — exact-type check (`isinstance(v, int) and not isinstance(v, bool) and v == 1`) with `{"result": True}` added to the unacknowledged matrix. Correct.
3. **README** — the v1r mode-table row and the "all local modes are read-only" overview are both updated to reflect authenticated local control and islanding's local-only, no-cloud-fallback semantics.

One note for the record: the hardware validation earlier in this thread was run at `70676f2`, and this commit changes runtime behavior in `gateway_manager.py` and `legacy.py` (the concurrency path itself). The earlier validation results should be considered superseded by this head — whether that warrants a quick re-run of the off-grid → on-grid sequence before merge is @jasonacox's call.

— Sam 🔌

jasonacox:
As part of this PR, should we add the "Go off grid" control to the "Powerwall Control" panel? If we did, I would want a confirmation / warning popup and a simple cooldown period to rate limit toggle attacks.

jasonacox-sam:
@jasonacox — I think yes, with exactly the two safeguards you named, plus a couple of implementation notes:

**Why it's worth adding:** the endpoint is authenticated, hardware-validated, and now hardened against the timeout race (a second toggle gets a 409 with "check grid status, don't retry"). Keeping it API-only means the only people who can use it safely are people who don't need the Console anyway. A guarded button makes it usable by the person standing in the garage watching the lights — which is actually the safest context for a manual grid transition ("test only when someone is present").

**Design suggestions:**
- **Confirmation modal** carrying the condensed safety warning (@mcmahonl19's expanded wording from above, trimmed for a modal): contactor operation, possible frequency rise while islanded, ack ≠ grid state, check `grid_status` after. A typed confirmation ("OFF GRID") is probably overkill for a modal, but a plain OK/Cancel is too light — a single deliberate checkbox ("I understand this will disconnect my home from the grid") plus the button feels right.
- **Cooldown enforced server-side, not just in JS.** A client-side cooldown does nothing against toggle attacks — the same `PW_RATE_LIMIT`-style thinking as #84. A simple per-command minimum interval enforced in the islanding route (independent of the in-flight 409) would rate-limit even scripted callers. I'd suggest something like 30–60s between opposite-direction islanding commands, returned as a 429 with retry-after.
- **State-aware button.** Fetch `grid_status` and render "Go off grid" vs "Reconnect grid" so the modal can state the *actual* current state rather than a guess — this also prevents the "send the opposite command after an unclear result" failure mode the README warns about.
- **Opt-in?** The PR deliberately shipped with no Console surface. A `--console-islanding` / env gate (default off, like the rate-limit middleware in #84) would let cautious users keep the Console as-is while early adopters get the button. Your call on whether that's warranted or over-cautious given the auth + confirm + cooldown stack.

Happy to spec this out as a follow-up to this PR (or a small extension of it, @mcmahonl19's call) once the base lands — the backend 409/cooldown piece and the Console piece could even split into two commits.

— Sam 🔌

mcmahonl19:
Agreed. I added a small initial Console surface to this PR, but you’re right that a browser-only cooldown is not a sufficient safety boundary. I’ll revise it to use server-side cooldown/409 with `Retry-After`, explicit Console opt-in, and a more deliberate confirmation flow. I’ll keep the backend and Console changes as separate commits so the safety boundary is easy to review.

mcmahonl19:
@jasonacox-sam That sounds helpful—please do. I’ve added a small initial Console surface so we have a concrete starting point, but I agree it should not define the safety model. Could you spec the expected server-side cooldown/response semantics, default-off opt-in, and confirmation/state behavior, and advise whether you’d prefer that as a small extension here or a follow-up after the base merges? I’ll hold the next revision until we’re aligned.

jasonacox-sam:
@tmp/pr104_reply.md

jasonacox:
I appreciate the warning text, but this makes the UI a bit too messy. I updated the UI:

# Old UI

image

# New UI

Minimized control into this:

image

With a popup confirmation modal:

image


image

jasonacox-sam:
**Recommendation: treat the current UI (`ff259dcc`) as the settled Console baseline and narrow the held revision to exactly the backend boundary — the server-side cooldown (429 + `Retry-After`) and the default-off Console opt-in.** Spec §3 (Console confirmation/state behavior) above is superseded by the new UI, which implements its substance more cleanly than my modal-and-checkbox wording did.

What the new UI gets right, checked against the spec:

- **State field + single action button** — this is the state-aware control §3 asked for, minus the clutter: only the action matching the reported state is offered, and the button is gated on control being enabled *and* v1r detection (`pw3` + `tedapi_mode === 'v1r'` from `/stats`). The refresh control next to the state field is the right "verify, don't assume" affordance.
- **Native `<dialog>` confirm** — the styled modal replaces `window.confirm`, and the inline hint carries the essential caveat without a wall of warning text. Cleaner and still deliberate.
- **One concession from the spec, and it's the right one:** §3 asked for a deliberate checkbox in the modal. With the server enforcing `confirm: true`, the cooldown, and the opt-in flag, the checkbox was friction theater — the boundary lives server-side, and dropping it keeps the modal honest.

So the remaining revision scope, @mcmahonl19, is §1–§2 of the spec, unchanged:

1. **Server-side cooldown** — per-gateway, `time.monotonic()` stamped at dispatch (not completion), both directions, and any prior outcome (200/502/503/still-running) starts the window; 429 rejections don't extend it. 60 s default with a `PW_ISLANDING_COOLDOWN` override clamped to a 30 s floor. Check order: in-flight first (409, no `Retry-After`), then cooldown (429 + integer `Retry-After`). Mirroring the existing `IslandingCommandInProgressError` pattern with a typed `IslandingCooldownError` keeps it trivially testable.
2. **Default-off opt-in** — `PW_CONSOLE_ISLANDING` gates only the Console surface, never the API endpoint; `/control/status` grows `"islanding": bool` (booleans only, since it's unauthenticated).
3. **UI wiring (small)** — on a 429, disable the action button for the `Retry-After` window and surface the wait. The client-side 60 s localStorage cooldown can stay as instant feedback but should always defer to the server's value, and the UI must never retry on its own after a timeout, matching the README guidance.

Commit split stays as planned: backend boundary first, Console wiring second.

The hardware re-run at the final head stays gated on the owner's go-ahead, as before — the validation record to date is as summarized upthread (the `70676f2` results were superseded by the concurrency changes at `4cf4bb4a`).

— Sam 🔌

jasonacox:
Pushed two commits wrapping up the remaining items — this is now feature-complete from my side.

## `b46ff6c` — release note wording
Aligned the Console entry in `RELEASE.md` with the final single-button UI.

## `bb597e6` — server-enforced islanding cooldown (`PW_ISLANDING_COOLDOWN`)

@jasonacox-sam's point about the browser-only cooldown was right, and I didn't want to merge with that gap or leave it to a follow-up — the 60s Console lockout is `localStorage` in one tab, so any client with the control token (curl, a buggy HA automation) could toggle the contactor as fast as the gateway acked. The cooldown is now enforced server-side:

- **`PW_ISLANDING_COOLDOWN`** — per-gateway, default **30s** (deliberately under the Console's 60s lock so UI users never hit it; it exists to catch API clients and automations), `0` disables.
- **Clock starts at dispatch and failures/timeouts count** — the gateway may have acted on an unacknowledged command, so it arms the cooldown too. Same principle as the rest of this PR: unknown outcome ≠ safe to retry.
- Inside the window → **HTTP 429 with `Retry-After`** and verify-first guidance; a command still in flight remains **409**. Check order: 401 → 400 → 409 → 429.
- Enforcement lives in `local_control()` (not the route), uses `time.monotonic()`, so no future caller can bypass it and wall-clock changes can't skew it.
- Console needed zero changes — a 429's message flows through the existing error path.
- 5 new tests (429+`Retry-After` bounds, cooldown-armed-by-failure, expiry, disabled-via-0, direct `local_control` enforcement). **378/378 passing** at `bb597e6`; README and `RELEASE.md` updated.

This implements the server-side cooldown from Sam's spec. The Console opt-in flag from that spec I'm deferring — the card already sits behind `PW_CONTROL_SECRET`, so it can be a follow-up if we still want it.

## Asks

@jasonacox-sam — please run a final hardware pass at `bb597e6`. Beyond re-running the original sequence, specifically exercise the cooldown: after the off-grid command, immediately send the reconnect via curl and confirm the **429 + `Retry-After`**, wait it out, then reconnect normally. Pin results to the SHA.

@mcmahonl19 — you'd offered to take this revision, and I didn't want to hold the merge on timezone alignment, so I went ahead — apologies for taking it out from under you, and thank you for the excellent foundation (the dispatch-time cooldown slot fell out naturally of your `_islanding_futures` structure). Please look over `bb597e6` and flag anything you'd do differently; happy to adjust before merge.

Once Sam's hardware pass is green and Lynn's had a look, I'll merge.

mcmahonl19:
Jason, thank you for carrying this across the finish line—and for taking the revision rather than letting my distractions hold it up. The final shape makes sense to me: a clean state-aware UI with a focused confirmation, fuller safety guidance in the documentation, and a server-side cooldown that protects every caller, not just the browser. It looks good from my side. I’ll leave it with Sam’s final hardware verification. Thanks again.

jasonacox-sam:
**Final hardware pass at `bb597e6` — green across the board ✅** (PW3 + Backup Gateway, TEDAPI v1r mode with registered key, test instance of the PR branch on a spare port so production monitoring was untouched)

Ran the original sequence plus the cooldown exercise exactly as requested — off-grid, immediate curl reconnect (rejected), wait out the window, reconnect normally:

| Step | Result |
|---|---|
| `GET /api/system_status/grid_status` baseline | 200 — `SystemGridConnected` |
| `POST /control/islanding {"value":"off_grid","confirm":true}` | 200 in 38 ms — `{"mode":6,"force":true,"result":1}` |
| **Immediate** `on_grid` (2 ms later) | **429** + `Retry-After: 30` — *"Islanding is rate limited; wait 30s and verify grid status before sending another command."* |
| Island confirmation | `SystemIslandedActive` after ~6 s |
| `off_grid` inside the window (both-directions check) | **429** + `Retry-After: 24` — window counts down, opposite direction equally blocked |
| Waited out the 30 s window | — |
| `POST /control/islanding {"value":"on_grid"}` | 200 in 49 ms — `{"mode":1,"force":false,"result":1}` |
| Reconnect confirmation | back to `SystemGridConnected` after ~18 s |

Negative probes re-verified at this SHA: wrong token → 401, `confirm: "true"` string → 400, `value: "bogus"` → 400, `off_grid` without `confirm` → 400. Final state verified `SystemGridConnected`; total islanded window ~57 s.

**A bonus the run produced for free:** my first launch accidentally used the system's pypowerwall 0.17.1 instead of the PR-pinned 0.17.3, so the off-grid dispatch *failed* (503, no islanding response — house never left the grid). Even that failure armed the cooldown: the immediately-following command got a clean 429 + `Retry-After: 30`. That's the "failures/timeouts count" clause of the spec confirmed on live hardware, not just in the test suite. (Side note worth having on record: 0.17.1's TEDAPI backend rejects `go_off_grid` outright — the `pypowerwall==0.17.3` pin in this PR is load-bearing, so the Docker image and any source installs must ship with it.)

**Review of `bb597e6` against the spec** — it implements §1 faithfully: `time.monotonic()` stamped at dispatch inside `local_control()` (no caller can bypass), typed `IslandingCooldownError` carrying the remaining seconds, mapped to 429 with an integer `Retry-After` in the route, 409 still reserved for in-flight, check order 401 → 400 → 409 → 429, non-islanding writes unaffected. The deltas from my spec are deliberate and I think correct: 30 s default (parked under the Console's 60 s lock so UI users never hit it — the boundary targets API clients) rather than my 60 s + 30 s floor, and the deferred Console opt-in (the card already sits behind `PW_CONTROL_SECRET`, so an extra flag can be a follow-up if still wanted).

Thank you @mcmahonl19 — the dispatch-time cooldown slot fell straight out of your `_islanding_futures` structure, and the foundation you built here (44 non-actuating tests arriving with the draft, the insistence on a straight validation record) is why this endpoint could be trusted to a live contactor at all. And the single-button Console UI with the server boundary under it is the right final shape.

From my side this is hardware-validated at `bb597e6` with the complete semantics — ack, state convergence, 409/429 contract, cooldown arm-on-failure — and ready for the merge call.

— Sam 🔌

jasonacox:
Thanks @mcmahonl19 ! This is a great addition to the server. 🙏 

@jasonacox-sam  Thanks for the hardware test! Please create a new tag for v0.6.5 and github release. Include all changes since last tag and any contributors. Post on X.

Container Name:  jasonacox/pypowerwall-server:0.6.5

View at:
https://pypi.org/project/pypowerwall-server/0.6.5/

jasonacox-sam:
@jasonacox — Done. Tag `v0.6.5` and the [GitHub release](https://github.com/jasonacox/pypowerwall-server/releases/tag/v0.6.5) are live.

Included everything since v0.6.4: the authenticated PW3 v1r islanding endpoint with server-enforced cooldown (`PW_ISLANDING_COOLDOWN`), the Console islanding controls with confirm dialog, and the pypowerwall 0.17.3 bump. Release notes credit **@mcmahonl19** for the design, implementation, and hardware verification, and @jasonacox for the Console UX polish and merge.

Confirmed PyPI 0.6.5 and container `jasonacox/pypowerwall-server:0.6.5` are live. Announced on X: https://x.com/jasonacox_sam/status/2098981343296393238

— Sam 🔌

PR #112, opened 2026-09-16
fix: show TEDAPI v1r (PW2) vs (PW3) by hardware

erikgieseler:
## Summary
GW2/PW2 with `v1r` (`PW_RSA_KEY_PATH`) shows `TEDAPI v1r (PW3)` and PW3 icon in `powerflow`, even though hardware is PW2.

## Root Cause
`pw3` is transport, not hardware:
- `app/core/gateway_manager.py:835` `data.pw3 = bool(pw.tedapi.pw3)` - `v1r` reports `True`
- `app/api/legacy.py:2145` `pw3 = any(data.pw3)` -> `/stats.pw3`
- `app/static/index.html:3041` `if(pw3) mode+=' (PW3)'` + `powerflow` asset choice

## Changes (minimal, 3 files + `powerflow`)
- `gateway_manager.py:832` - hardware via `tedapi_config` `battery_blocks[].type` (`Powerwall3`/`LFPV`), fallback to last cached config, else leave `pw3=None` -> Health shows generic `TEDAPI v1r` for 1 poll instead of flash
- `index.html:3041` - `v1r` -> `(PW3)` if `true`, `(PW2)` if `false`, generic if `None`
- `powerflow/app.js:21343` - default to PW3 asset (`null` -> PW3, only `false` -> PW2)
- `RELEASE.md` `[0.6.7] - Upcoming` Fixed entry

## Validation
- PW2 `v1r` now `TEDAPI v1r (PW2)` + PW2 model in `Powerwall Status` (verified on PW2)
- `py_compile` + `node --check` clean
- No new config, no version bump

> ⚠️ **Testing Note - help needed:** Islanding is PW3 `v1r`-only per `pypowerwall` (`go_off_grid` requires `v1r`). On PW2 `v1r` the panel stays hidden by design - **I can’t test PW2 islanding currently and have no PW3 hardware**, would appreciate someone with PW3 `v1r` confirming `TEDAPI v1r (PW3)` still shows correctly. If `go_off_grid` is later verified working on PW2, the hardware check can be relaxed to enable it there as well.


erikgieseler:
@jasonacox & @jasonacox-sam: this PR is ready for review.

jasonacox-sam (commented):
**Verdict:** Solid, well-layered fix — verified against upstream semantics and the full test suite. Ready from my side; merge is the owner's call.

Erik, thank you for this — the root-cause analysis in the PR description is exactly right, and I especially appreciate that you validated it on real PW2 v1r hardware and then explicitly flagged the PW3 side as untested instead of hand-waving it. Also nice work isolating the fix after unwinding the MQTT experiment — the final net diff is clean and minimal.

**What I verified:**

- **Root cause confirmed upstream** — pypowerwall's `tedapi.pw3` is transport-based: `_connect_v1r()` sets it to `True` unconditionally (`v1r is PW3-only`), so a PW2 system on the v1r transport gets mislabeled as PW3. Deriving hardware at the server layer is the right call.
- **Hardware markers match existing conventions** — `"Powerwall3" in type or type == "LFPV"` mirrors the Console's existing model mapping, and the `1707000-*` part-number prefix matches the PW3 DIN prefix seen in pypowerwall's own fixtures. The `PackagePartNumber`/`partNumber`/`PartNumber` fallback ordering is sensible.
- **Tri-state handling is sound** — `None` (unknown) survives to `/stats` instead of being coerced to `false`, non-v1r connections keep the transport flag when config is unavailable, and the fallback to the last cached `tedapi_config` prevents label flapping on transient config-read failures. That addresses all three points from the earlier automated review, with a regression test for each.
- **Frontend is null-safe** — the islanding gate uses strict `=== true`, and the powerflow asset defaulting to PW3 while hardware is unknown self-corrects within one poll.
- **Tests** — the 5 new regression tests pass, and the full suite passes (403 passed). Cold start (null), known-PW2, PW3-via-type, PW3-via-part-number, and non-v1r transport preservation are each covered.

**Non-blocking nits** (fine for a follow-up):

1. `gateway_manager.py` (~line 895): the final `elif _raw_pw3_status is not None:` branch is unreachable — reaching it would require `_raw_pw3_status` to be non-`None` after the preceding branch already tested exactly that. Harmless, but it reads like it does something and could be dropped.
2. The part-number check reads `PackagePartNumber`/`partNumber`/`PartNumber`, but config `battery_blocks` in the wild commonly carry the part number inside the `vin` field (`<part>--<serial>`), which the battery-blocks code elsewhere already parses. Adding a `vin.split("--")[0].startswith("1707000")` check to that fallback would make it more robust — the type check covers most real systems today, so this is belt-and-suspenders.
3. Behavior note worth keeping in mind for release notes: on PW3 v1r, the islanding control now appears one poll later than before (it is gated on `pw3 === true`, which is null during cold start). Self-corrects; just a small timing change.

And +1 to the testing note in the description — the PW3 v1r path is unit-tested but not yet confirmed on real hardware. If anyone in the community runs a PW3 with `PW_RSA_KEY_PATH` configured, a quick confirmation that Health shows `TEDAPI v1r (PW3)` would close the loop.

— Sam 🔌

erikgieseler:
@jasonacox-sam dead branch removed, vin fallback + test, 404 passed

jasonacox-sam:
Confirmed against the head (`3bb8b5b`) — the unreachable `elif` branch is gone, the `vin` fallback is in (`split("--")[0]` after the explicit part-number fields), and `test_v1r_hardware_pw3_via_vin_part_number()` covers it, alongside the other five regression tests. Suite count checks out (409 = 403 + 6). That closes both non-blocking nits — ready from my side; merge is Jason's call.

— Sam 🔌

